PR #62, opened 2021-01-20
Add Sparkle

interstateone:
- Added [Sparkle](http://sparkle-project.org) via SPM
- Added Check for updates... menu item
- Added [Preferences](https://github.com/sindresorhus/Preferences) via SPM for toolbar item-style tabbed preferences window
- Split up the existing preferences into General and Advanced panes using Preferences, replacing the existing SwiftUI.Settings scene
- Added new Updates pane for Sparkle preferences
  - Includes check automatically, include prerelease preferences, last checked date and check now button
- Updated release steps to fix a .zip issue
- Updated release steps and AppCast/ to include Sparkle signature in appcast
- Added section to DECISIONS.md

The transitions between preference panes aren't great because of how the SwiftUI views layout the first time, but I think this tabbed preference window style is more idiomatic.

Updating is not _entirely_ working yet, because while this branch is behind the latest release version (1.0.1 (2)), and so it thinks it can update to it, the latest release version is from main and so doesn't have the EdDSA public key in its Info.plist and so updating fails.

|General|Updates|Advanced|
|-----|-----|-----|
|Screen Shot 2021-01-19 at 8 48 30 PM|Screen Shot 2021-01-19 at 8 48 33 PM|Screen Shot 2021-01-19 at 8 48 35 PM|

|up to date|new version|
|-----|-----|
| Screen Shot 2021-01-19 at 10 13 33 PM | Screen Shot 2021-01-19 at 10 42 28 PM |


csykes (approved):
Code looks good and functions as expected.

The right hand side feels a bit squishy if you felt like giving it a bit more room...
Screen Shot 2021-01-20 at 8 33 46 AM

PR #75, opened 2021-01-21
Update issue templates

MattKiazyk:
Adds some issue templates for ease of use


interstateone (approved):
Good idea, thanks Matt. We should probably add these to xcodes too.

PR #78, opened 2021-01-23
#71 Update Filter/Info to use AppStorage

### LINKED ISSUE #71: Filter/info pane doesn't persist across app closes
Issue:
If I tap on the filter or the info pane, it should persist across app loads

image

Repro Steps:
1. Set the filter on, or the info pane on
2. Close the app, reopen
3. App shows all versions again

This _was_ working previously.

MattKiazyk:
Closes #71 

Update the Filter and info button to use Appstorage instead of SceneStorage.

Only download I really see now, is in the multiple window situation, updating one, automatically updates the others.

### DIFF
--- a/Xcodes/Frontend/MainWindow.swift
+++ b/Xcodes/Frontend/MainWindow.swift
@@ -6,8 +6,8 @@ struct MainWindow: View {
     @State private var selectedXcodeID: Xcode.ID?
     @State private var searchText: String = ""
     @AppStorage("lastUpdated") private var lastUpdated: Double?
-    @SceneStorage("isShowingInfoPane") private var isShowingInfoPane = false
-    @SceneStorage("xcodeListCategory") private var category: XcodeListCategory = .all
+    @AppStorage("isShowingInfoPane") private var isShowingInfoPane = false
+    @AppStorage("xcodeListCategory") private var category: XcodeListCategory = .all
   
     var body: some View {
         HSplitView {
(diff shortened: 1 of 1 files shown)


Xcodes/Frontend/MainWindow.swift
```diff
@@ -6,8 +6,8 @@ struct MainWindow: View {
     @State private var selectedXcodeID: Xcode.ID?
     @State private var searchText: String = ""
     @AppStorage("lastUpdated") private var lastUpdated: Double?
-    @SceneStorage("isShowingInfoPane") private var isShowingInfoPane = false
-    @SceneStorage("xcodeListCategory") private var category: XcodeListCategory = .all
+    @AppStorage("isShowingInfoPane") private var isShowingInfoPane = false
+    @AppStorage("xcodeListCategory") private var category: XcodeListCategory = .all
```
interstateone, line 10: 👌 this is good, let's add a comment explaining why this is the way it is so that we can (hopefully) revert this later when the behaviour is changed. I just submitted a feedback report for this.

```suggestion
    // These two properties should be per-scene state managed by @SceneStorage property wrappers.
    // There's currently a bug with @SceneStorage on macOS, though, where quitting the app will discard the values, which removes a lot of its utility.
    // In the meantime, we're using @AppStorage so that persistence and state restoration works, even though it's not per-scene.
    // FB8979533 SceneStorage doesn't restore value after app is quit by user
    @AppStorage("isShowingInfoPane") private var isShowingInfoPane = false
    @AppStorage("xcodeListCategory") private var category: XcodeListCategory = .all
```

interstateone (approved):

PR #79, opened 2021-01-23
Catch and reword xip "not enough free space" error

### LINKED ISSUE #77: Catch and reword xip's "not enough free space" error

When the volume doesn't have enough space to expand a .xip archive, the `xip` command used under the hood will fail.

From the logs:

> 2021-01-23 11:16:57.291081-0700 Xcodes[17671:962405] [subprocess] Process.run error: xip: error: The archive “Xcode-12.4.0-Release.Candidate+12D4e.xip” can’t be expanded because the selected volume doesn’t have enough free space.

This is the alert that is currently shown:

Screen Shot 2021-01-23 at 11 17 25 AM

This is one of the most common errors I get when installing an Xcode.

Xcodes.app should present an alert with a good user-facing error message, like:

> The archive “Xcode-12.4.0-Release.Candidate+12D4e.xip” can’t be expanded because the current volume doesn’t have enough free space.

interstateone:
This change catches the xip error that is output when there's not enough free space to expand a .xip archive and makes sure that it's presented with a better error message.

It also adds AppStateTests to the XcodesTests target, not sure what happened there.

|Before|After|
|-----|-----|
|Screen Shot 2021-01-23 at 11 17 25 AM|Screen Shot 2021-01-23 at 11 32 41 AM|

Closes #77


MattKiazyk:
Language looks ok.. I have plenty of space so can't try it out :)

MattKiazyk (approved):

PR #80, opened 2021-01-23
Handle helper installation errors

### LINKED ISSUE #69: Install helper error shown multiple times after cancel install
Issue: 
If I hit cancel on the privileged helper I shouldn't be prompted multiple times saying it couldn't communicate with helper app. It should just cancel

RepoSteps:

1. Tapped `Make Active` 
2. Hit cancel on the prompt

image
3. Prompted multiple times with error 
image

This may be resolved with changes in #66

interstateone:
Instead of ignoring errors (either cancellation or failure) from helper installation, this change throws them so callers can react appropriately. Previously, whatever was being done that required the helper would proceed as if installation succeeded, which would always fail. Now, cancellation or failures will complete the subscriptions without attempting the next steps, and without showing more errors to the user. The initial error, for example cancellation in the below screenshot, will still be shown.

The second commit makes a small change to how auth error messages are created to avoid describing the optional wrapper type with `Optional(...)`.

|after cancelling helper installation|
|-----|
|Screen Shot 2021-01-23 at 1 51 35 PM|

Closes #69

### DIFF
--- a/Xcodes/Backend/AppState+Install.swift
+++ b/Xcodes/Backend/AppState+Install.swift
@@ -303,11 +303,10 @@ extension AppState {
     }
 
     func enableDeveloperMode() -> AnyPublisher<Void, Error> {
-        if helperInstallState == .notInstalled {
-            installHelper()
-        }
-
-        return Current.helper.devToolsSecurityEnable()
+        installHelperIfNecessary()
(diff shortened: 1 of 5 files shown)


MattKiazyk (approved):
Seems to run as advertised

PR #81, opened 2021-01-23
Show ToolTip on progress indicator

MattKiazyk:
I didn't like not knowing what percent my download was at, so a quick win is to show it on the tooltip

image

### DIFF
--- a/Xcodes/Frontend/Common/ProgressIndicator.swift
+++ b/Xcodes/Frontend/Common/ProgressIndicator.swift
@@ -23,7 +23,6 @@ struct ProgressIndicator: NSViewRepresentable {
         nsView.controlSize = controlSize
         nsView.isIndeterminate = isIndeterminate
         nsView.style = style
-        nsView.toolTip = "Dowloading: \(Int(doubleValue * 100.0))% complete"
     }
 }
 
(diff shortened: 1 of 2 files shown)


Xcodes/Frontend/Common/ProgressIndicator.swift
```diff
@@ -23,6 +23,7 @@ struct ProgressIndicator: NSViewRepresentable {
         nsView.controlSize = controlSize
         nsView.isIndeterminate = isIndeterminate
         nsView.style = style
+        nsView.toolTip = "Dowloading: \(Int(doubleValue * 100.0))% complete"
```
interstateone, line 26: This view is only used in the one spot right now, but I'd like to keep it as generic as possible. Could this tooltip be added with a `.help()` modifier [at the call site in InstallationStepView](https://github.com/RobotsAndPencils/XcodesApp/blob/main/Xcodes/Frontend/XcodeList/InstallationStepView.swift#L19) instead?

interstateone (approved):
Looks like a good addition, just the one comment about where to add this in code. Thanks Matt!

MattKiazyk:
Moved.. sorry, i swear I tried there and it had errors about not being able to find `.help` which I took as it not being compatible with a UIKit progress

PR #84, opened 2021-01-24
Prepare user for helper installation before attempting to install

### LINKED ISSUE #66: Prepare the user for helper installation alert before attempting it
This would be similar to what a lot of mobile apps do for location services permission. I'm imagining this will look something like: an alert that we show first, explaining what the helper is and why it's necessary, and only if they consent would we attempt to install the helper. If they decline then we don't attempt the install. The copy in Preferences > Advanced is probably close to what we'd show in the alert.

interstateone:
The intent of this change is to make sure users are prepared to enter their macOS account password in the helper installation alert so that they aren't surprised and cancel installation, which leads to further failures.

This affects making Xcode versions active and the "step 6: finishing" post-install steps of installation.

The implementation of these changes aren't straightforward (I think), as most of the logic is implemented in publishers that now have to wait on user interaction in the preparation alert. This is done in two different ways because the way these methods are invoked is a little different:

- `select(id:shouldPrepareUserForHelperInstallation:)` is invoked once and returns early to prepare the user. Once the user makes a decision in the preparation alert the method is automatically invoked a second time to perform the action if they consented to installation. If they didn't consent then nothing happens. They can just try to make the version active again.
- `performPostInstallSteps(for:)` is invoked only once as part of a larger installation publisher. It returns one of two publishers. If the helper is not already installed, it'll return a publisher that waits for the user to make a decision in the preparation alert before continuing with the post-install steps. If the helper is already installed, the method will return a publisher that only performs the post-install steps. If the user doesn't consent to installation, the publisher fails so that they can be informed what the result of installation with the post-install steps being performed is. This is different from `select` because there is no "just try clicking it again" option in this case.

The user is be prepared and prompted for installation only once for all three post-install steps.

Because the post-install steps are optional as far as Xcodes.app is concerned (the user will just have to perform them manually when they launch the new Xcode), an alert is shown with an error message explaining what happened, but this doesn't cause installation to fail.

---

## Testing

- Install an Xcode and when the preparation alert is shown, cancel. You should see an error about the post-install steps not being performed, but the new Xcode should appear installed. (The button on the right should read "Open" instead of "Install")
- Install an Xcode and when the preparation alert is shown, click Install, but then cancel the helper installation alert. You should see an error about the post-install steps not being performed, but the new Xcode should appear installed.
- Install an Xcode and when the preparation alert is shown, click Install, and then enter your macOS account password to install the helper. You should see no errors and the new Xcode should appear installed.
- Clicking the alert buttons in different situations, particularly when the window is inactive, should always perform the button's action, and may present an alert twice, but should never crash. I don't know why the double alert thing is happening but I think this is the best combination of those three things for now.

I also added a "Perform post-install steps" context menu item to the list row context menu which is only shown in debug builds.

|Preparation alert|Post-install failed with helper|Post-install failed without helper|
|-----|-----|-----|
|Screen Shot 2021-01-23 at 7 56 37 PM|Screen Shot 2021-01-23 at 3 24 54 PM|Screen Shot 2021-01-23 at 3 31 24 PM|

---

Closes #66

### DIFF
--- a/Xcodes/Frontend/XcodeList/XcodeListViewRow.swift
+++ b/Xcodes/Frontend/XcodeList/XcodeListViewRow.swift
@@ -37,13 +37,20 @@ struct XcodeListViewRow: View {
                 InstallButton(xcode: xcode)
             case .installing:
                 CancelInstallButton(xcode: xcode)
-            case .installed:
+            case let .installed(path):
                 SelectButton(xcode: xcode)
                 OpenButton(xcode: xcode)
                 RevealButton(xcode: xcode)
                 CopyPathButton(xcode: xcode)
                 Divider()
                 UninstallButton(xcode: xcode)
+                
+                #if DEBUG
+                Divider()
+                Button("Perform post-install steps") {
+                    appState.performPostInstallSteps(for: InstalledXcode(path: path)!) as Void
+                }
+                #endif
             }
         }
     }
--- a/Xcodes.xcodeproj/project.pbxproj
+++ b/Xcodes.xcodeproj/project.pbxproj
@@ -80,6 +80,7 @@
 		CAC281DA259F985100B8AB0B /* InstallationStep.swift in Sources */ = {isa = PBXBuildFile; fileRef = CAC281D9259F985100B8AB0B /* InstallationStep.swift */; };
 		CAC281E2259FA44600B8AB0B /* Bundle+XcodesTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = CAC281E1259FA44600B8AB0B /* Bundle+XcodesTests.swift */; };
 		CAC281E7259FA45A00B8AB0B /* Environment+Mock.swift in Sources */ = {isa = PBXBuildFile; fileRef = CAC281E6259FA45A00B8AB0B /* Environment+Mock.swift */; };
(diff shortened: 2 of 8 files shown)


Xcodes/Frontend/XcodeList/XcodeListViewRow.swift
```diff
@@ -37,13 +37,20 @@ struct XcodeListViewRow: View {
...
                 SelectButton(xcode: xcode)
                 OpenButton(xcode: xcode)
                 RevealButton(xcode: xcode)
                 CopyPathButton(xcode: xcode)
                 Divider()
                 UninstallButton(xcode: xcode)
+                
+                #if DEBUG
+                Divider()
+                Button("Perform post-install steps") {
```
interstateone, line 50: I figured this couldn't hurt to keep in the code behind the debug flag... It's nice to have for testing during development instead of trying to go through the full install flow all the time.

MattKiazyk (approved):
This works as advertised. My only comment would be on the post install states, perhaps adding an action on the alert to `Launch xcode X.X` since you're already indicating that is the only way to proceed.

interstateone:
👍 Good idea, I think I need to make a change to the ErrorHandling library to support that first.

PR #85, opened 2021-01-25
Add progress indicator on dock icon

### LINKED ISSUE #17: Display aggregate installation progress in dock icon
At a minimum, Xcodes should display an appropriate overlay on the dock icon when one or more versions are being installed. This could be modelled as a boolean internally.

A fancier implementation of this could provide richer indication of the aggregate installation progress, perhaps through a tree of Progress objects for all versions that are being installed, and each step within their installation.

A library that might make this easier to implement is https://github.com/sindresorhus/DockProgress

senmu:
Closes #17

Uses suggested [DockProgress](https://github.com/sindresorhus/DockProgress) to display a progress bar overlay on top of the dock icon as an Xcode is downloading.

Screen Shot 2021-01-24 at 10 41 38 PM

### DIFF
--- a/Xcodes/Frontend/XcodeList/InstallationStepView.swift
+++ b/Xcodes/Frontend/XcodeList/InstallationStepView.swift
@@ -17,7 +17,7 @@ struct InstallationStepView: View {
                     controlSize: .small,
                     style: .spinning
                 )
-                .help("Dowloading: \(Int((progress.fractionCompleted * 100)))% complete")
+                .help("Downloading: \(Int((progress.fractionCompleted * 100)))% complete")
             case .unarchiving, .moving, .trashingArchive, .checkingSecurity, .finishing:
                 ProgressView()
                     .scaleEffect(0.5)
--- a/Xcodes.xcodeproj/project.pbxproj
+++ b/Xcodes.xcodeproj/project.pbxproj
@@ -99,6 +99,7 @@
 		CAFE4AB425B7D3AF0064FE51 /* AdvancedPreferencePane.swift in Sources */ = {isa = PBXBuildFile; fileRef = CAFE4AB325B7D3AF0064FE51 /* AdvancedPreferencePane.swift */; };
 		CAFE4ABC25B7D54B0064FE51 /* UpdatesPreferencePane.swift in Sources */ = {isa = PBXBuildFile; fileRef = CAFE4ABB25B7D54B0064FE51 /* UpdatesPreferencePane.swift */; };
 		CAFFFED8259CDA5000903F81 /* XcodeListViewRow.swift in Sources */ = {isa = PBXBuildFile; fileRef = CAFFFED7259CDA5000903F81 /* XcodeListViewRow.swift */; };
+		E689540325BE8C64000EBCEA /* DockProgress in Frameworks */ = {isa = PBXBuildFile; productRef = E689540225BE8C64000EBCEA /* DockProgress */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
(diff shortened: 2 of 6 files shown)


Xcodes/Frontend/XcodeList/InstallationStepView.swift
```diff
@@ -17,7 +17,7 @@ struct InstallationStepView: View {
                     controlSize: .small,
                     style: .spinning
                 )
-                .help("Dowloading: \(Int((progress.fractionCompleted * 100)))% complete")
+                .help("Downloading: \(Int((progress.fractionCompleted * 100)))% complete")
```
MattKiazyk, line 20: 🙇

MattKiazyk (commented):
This worked for me. Tried cancelling and restarting while it was downloading. All reset itself.

I'll leave Brandon for final :shipit:  but looks good

interstateone:
Nice! Thanks Sam.

One thing I'm not sure about is if and how this should represent indeterminate InstallationSteps (all of the ones other than downloading, for example unarchiving can take longer than downloading to complete but we don't currently show determinate progress for it) and multiple simultaneous installations. Perhaps we could represent both of these with a "richer" use of Progress, like an instance per installation step, with a parent Progress for the entire installation, and a parent Progress of that to track aggregate installation progress. There might be other options too.

What are your thoughts on this?

MattKiazyk:
My 2 cents: Create another enhancement ticket and get this out (MVP). Feel like there's better UX we could do for that unxip step that we should think about better.

senmu:
I think it would be good to investigate getting the dock progress to also represent those other steps, but I agree with Matt as implementing that seems to be a hurdle.

The UX would indeed be improved because at the moment, when the progress indicator completes, a user would go back to the app and feel slight disappointment as those remaining steps can take a while to complete. However, in its present form, it still makes the experience better than it was without it because the only way to tell how progress was going is to keep the app in front.

If you're alright with that course of action, I'll create a follow up enhancement to this so that the dock progress can report from the start of the Xcode download to when it's fully ready to use.

interstateone:
Thanks guys. Agreed that the tree-of-Progress approach would be a fair bit more work, and might not be needed for a minimal version of this feature, so I'll take a step back from that particular approach. I'll also note that the issue I created was worded "aggregate download progress" but probably should have been "aggregate installation progress," so I apologize for that.

What's proposed here is to have the dock show "nothing" or "some progress", which I think most people will interpret as "installing something" or "not installing anything." This will be misleading though, because the dock icon showing nothing is actually representing either "not installing anything" or "installing something, but not the downloading part." Users will still need to rely on keeping the app in front to know when installation is complete.

I think a minimum version of this feature could be something like "represent a Bool in the dock icon", where the Bool is "one or more versions have an installation state of `installing`," and the representation is something appropriate for indeterminate progress, like a spinner (doesn't come for free, and based on [this](https://github.com/sindresorhus/DockProgress/issues/1) would take some under-the-hood work), a static down arrow, etc.

---

> Feel like there's better UX we could do for that unxip step that we should think about better.

Agreed, and I think that (specifically, getting percent complete via private API) could be a valuable change, but I think representing that this step is happening is valuable even without knowing the percent complete because it takes so long.

MattKiazyk:
Could do a simple badge animation? 

Jan-25-2021 14-34-50

MattKiazyk:
Or another example: 
Jan-25-2021 14-45-43

interstateone:
I'd prefer to draw an overlay on the icon tile (like how DockProgress is implemented) only because it frees up the badge for other things that are more appropriate, like count of completed installations. I'd like to explore the other options, and if animated badge content is the easiest thing to do for now then that's okay.

How are you driving the character animations, and could that be used to draw those in an overlay on the dock tile instead of in the badge?

MattKiazyk:
that is just ascii animations, so just a Timer + looping through an array of text

senmu:
Closing in favour of #423

PR #86, opened 2021-01-26
Improve authentication error messages that are shown to users

interstateone:
This PR makes two changes:

Improve how auth errors are shown in an Alert. They all provide their own descriptions now, where before some of them would default to String(describing:) which didn't do a great job. I also removed some of the associated values from the descriptions because while it might be useful for debugging, it is not immediately helpful for users. We may want to find another way to provide users with this information, even if it's how to find it in Console.app, in the future. A 400 status code during security code submission is now treated as an "incorrect code" error. I'm not sure why it wasn't before, as I thought this implementation was just taken from xcodes, but when I tested this situation I got a 400 instead of a 401.

Added bug report and feature request items to the Help menu so it's easier for users to find these options and so that we can refer to them in error messages.

### DIFF
--- a/Xcodes/AppleAPI/Sources/AppleAPI/Client.swift
+++ b/Xcodes/AppleAPI/Sources/AppleAPI/Client.swift
@@ -129,7 +129,7 @@ public class Client {
                     case 400, 401:
                         throw AuthenticationError.incorrectSecurityCode
                     case let code:
-                        throw AuthenticationError.badStatusCode(code, data, urlResponse)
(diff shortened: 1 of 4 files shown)


MattKiazyk (approved):
I tried a few different ways to error and work as advertised. 

I did in my testing find a new bug which is created in  #88

interstateone:
Interesting, good catch Matt. Looking again at the Preferences library I see what I did wrong.

PR #89, opened 2021-01-27
Updates PreferencesView To use built in Settings Scene

### LINKED ISSUE #88: Preferences Window is shown multiple times
**Describe the bug**
The preferences window can be created multiple times. That isn't that bad amongst itself, however when you start presenting alerts on it, you can end up with multiple windows showing multiple alerts. 

image

**To Reproduce**
1. Go to menu and open up preferences
2. Go back to main view and open up a open preferences from step 1

You now have multiple preferences windows

**Expected behavior**
I would expect only 1 preference window to be always shown. If the window is a presentView (ie I have to close it before going back to the main window) I think that's ok.

MattKiazyk:
Fixes #88 

Uses SwiftUI's Settings Scene - https://developer.apple.com/documentation/swiftui/settings Instead of PreferencesWindowController that comes with the `Preferences` add in.

Preferences dependency has been removed. Converted the style with a `PreferencesGroupBoxStyle`, which turned out pretty well. 

Did a quick test for the options and things seems to work. Some awkwardness with InstallHelper as it closes the preferences window automatically, but it works.

image

The only annoyance that I couldn't figure out is the advanced view looks to have a different background color, compared to the other views, but I couldn't figure out why.

### DIFF
--- a/Xcodes/Frontend/Preferences/AdvancedPreferencePane.swift
+++ b/Xcodes/Frontend/Preferences/AdvancedPreferencePane.swift
@@ -8,7 +8,7 @@ struct AdvancedPreferencePane: View {
     
     var body: some View {
         VStack(alignment: .leading, spacing: 20) {
-            GroupBox(label: Text("DataSource")) {
+            GroupBox(label: Text("Data Source")) {
                 VStack(alignment: .leading) {
                     Picker("Data Source", selection: $dataSource) {
                         ForEach(DataSource.allCases) { dataSource in
@@ -66,8 +66,7 @@ struct AdvancedPreferencePane: View {
             }
             .groupBoxStyle(PreferencesGroupBoxStyle())
         }
-        .padding()
-        .frame(width: 500)
+        .frame(width: 400)
     }
     
     private var dataSourceFootnote: NSAttributedString {
--- a/Xcodes/Frontend/Preferences/AdvancedPreferencePane.swift
+++ b/Xcodes/Frontend/Preferences/AdvancedPreferencePane.swift
@@ -2,10 +2,6 @@ import AppleAPI
 import Preferences
 import SwiftUI
 
-extension Preferences.PaneIdentifier {
-    static let advanced = Self("advanced")
-}
-
 struct AdvancedPreferencePane: View {
     @EnvironmentObject var appState: AppState
     @AppStorage("dataSource") var dataSource: DataSource = .xcodeReleases
--- a/Xcodes/Frontend/Preferences/AdvancedPreferencePane.swift
+++ b/Xcodes/Frontend/Preferences/AdvancedPreferencePane.swift
@@ -1,5 +1,4 @@
 import AppleAPI
-import Preferences
 import SwiftUI
 
 struct AdvancedPreferencePane: View {
@@ -8,8 +7,8 @@ struct AdvancedPreferencePane: View {
     @AppStorage("downloader") var downloader: Downloader = .aria2
     
     var body: some View {
-        Preferences.Container(contentWidth: 400.0) {
-            Preferences.Section(title: "Data Source") {
+        VStack(alignment: .leading, spacing: 20) {
+            GroupBox(label: Text("DataSource")) {
                 VStack(alignment: .leading) {
                     Picker("Data Source", selection: $dataSource) {
                         ForEach(DataSource.allCases) { dataSource in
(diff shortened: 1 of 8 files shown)


MattKiazyk:
On occasion, the advanced tab has trouble rendering:

image

My guess is has to do with the attributedString texts.

Xcodes/Frontend/Preferences/AdvancedPreferencePane.swift
```diff
@@ -1,19 +1,14 @@
...
 struct AdvancedPreferencePane: View {
     @EnvironmentObject var appState: AppState
     @AppStorage("dataSource") var dataSource: DataSource = .xcodeReleases
     @AppStorage("downloader") var downloader: Downloader = .aria2
     
     var body: some View {
-        Preferences.Container(contentWidth: 400.0) {
-            Preferences.Section(title: "Data Source") {
+        VStack(alignment: .leading, spacing: 20) {
+            GroupBox(label: Text("DataSource")) {
```
interstateone, line 11:
```suggestion
            GroupBox(label: Text("Data Source")) {
```

Xcodes/Frontend/Preferences/AdvancedPreferencePane.swift
```diff
@@ -114,3 +113,20 @@ struct AdvancedPreferencePane_Previews: PreviewProvider {
         }
     }
 }
+
+// A group style for the preferences
+struct PreferencesGroupBoxStyle: GroupBoxStyle {
```
interstateone, line 118: 👌 this is a great use of GroupBox

Xcodes/Frontend/Preferences/AdvancedPreferencePane.swift
```diff
@@ -67,8 +64,10 @@ struct AdvancedPreferencePane: View {
                     Spacer()
                 }
             }
+            .groupBoxStyle(PreferencesGroupBoxStyle())
         }
-        .padding(.trailing)
+        .padding()
+        .frame(width: 500)
```
interstateone, line 70: I find there's a bit too much padding in this window because each pane is adding padding, and PreferencesView is also adding some. Since these two lines are the same in all of the panes, you could probably use them once in PreferencesView instead, which would also reduce the padding.

interstateone (commented):
Nice work Matt! This looks good, it's nice to be able to do this in SwiftUI without another library.

Can you elaborate on the helper installation closing the preferences window? I tried it out and it changed the main window to be key because that's where the helper alerts are presented from, but the preferences window was still open. I think that's how it worked before too, and so if that's what you were seeing I think that's okay.

Regarding the layout issues in the advanced pane, I didn't see that when I was testing. Were there particular steps that reproduced that for you? I think you're right about it being the attributed string stuff.

I'd also expect a change to Licenses.rtf because you removed Preferences. It's possible that the build phase isn't running at the right times. Could you try cleaning and building the project to see if that changes it?

Xcodes/Frontend/Preferences/AdvancedPreferencePane.swift
MattKiazyk: Removed some extra padding and set to 400 width.. Think it looks pretty good.

MattKiazyk:
> Can you elaborate on the helper installation closing the preferences window? I tried it out and it changed the main window to be key because that's where the helper alerts are presented from, but the preferences window was still open. I think that's how it worked before too, and so if that's what you were seeing I think that's okay.

Yeah exactly that.. was a bit jaring for the settings window to close suddenly, instead of being prompted inside the settings view.

> Regarding the layout issues in the advanced pane, I didn't see that when I was testing. Were there particular steps that reproduced that for you? I think you're right about it being the attributed string stuff.

Just randomly came up by switching back and forth between tabs. 

> I'd also expect a change to Licenses.rtf because you removed Preferences. It's possible that the build phase isn't running at the right times. Could you try cleaning and building the project to see if that changes it?

What's weird is it isn't removed. What's adding to the problem is ever since I was on the iconProgress branch, all the license.rtf ever does (even in this branch) is try to add the line for `DockProgress`

image

MattKiazyk:
File > Swift Packages > Reset Package Cache resolved the issue with license.rtf

Xcodes/Frontend/Preferences/PreferencesView.swift
```diff
@@ -0,0 +1,32 @@
...
+                .tag(Tabs.updates)
+            AdvancedPreferencePane()
+                .environmentObject(appState)
+                .tabItem {
+                    Label("Advanced", systemImage: "gearshape.2")
+                }
+                .tag(Tabs.advanced)
+        }
+        .padding(20)
+        .background(Color(.textBackgroundColor)) // sets the background for all tabs to be the same.
```
interstateone, line 30: I'm not sure that this is necessary, and it might be the wrong semantic color for this situation. Was there a reason that you added it?

Xcodes/Frontend/Preferences/AdvancedPreferencePane.swift
interstateone: Agreed, thanks Matt

interstateone (approved):

interstateone:
Project builds and tests pass locally, so I'll merge ahead of the GitHub Actions Big Sur Queue™.

Issue #94, opened 2021-01-31
Display File Size in Info panel


kurisubrooks:
**Tell us how we can improve Xcodes**
Ideally, I'd like to be able to see the size of the Xcode release before downloading it (as I have a fairly limited data connection, so this information is important to me.)

**Is your feature request related to a problem? Please describe.**
No.

**What would you like to see? How would you like it to work?**
In the info panel, upon clicking a release, display the file size of the release. Just text is fine, no need to be fancy.

MattKiazyk:
Hi @kurisubrooks 

If you install the latest version of xcodes you'll see the download size is now in the info panel when using the Apple data source. When using the xcodereleases, it unfortunately is not, as that requires some changes to a 3rd party (open source) tool which I've created a PR for that will hopefully be accepted. 

Furthermore however, once you start downloading, the file size does show up in either data source. Hopefully that will help you. 

I'm going to close this issue for now. If you have any new issues - please feel free to create a new issue. 

Thanks

PR #98, opened 2021-02-02
Adds a DownloadFileSize when using Apple Data Source

MattKiazyk:
Connects #94 (Partially addressed)

When using the Apple datasource, this data is easily available. Added to info pain under a `Download Size` header. It will hide when the version is downloaded, as it is kind of irrelevant information.

Added a quick bytes to readable string converter. Just grabbed a quick one from the internet. 

image

XcodeReleases.com doesn't have this information. In order to fully support it with both datasources, we should possibly PR it for them.


Xcodes/Backend/Downloads.swift
```diff
@@ -13,5 +13,6 @@ public struct Download: Codable {
 
     public struct File: Codable {
         public let remotePath: String
+        public let fileSize: Double
```
interstateone, line 16: I think it would help to have the unit of this measurement be more clear for each of these new byte count properties, with names like `fileSizeInBytes`, or a type like `typealias ByteCount = Double`, etc.

interstateone (approved):
This works well Matt! I left a few comments but nothing major, I'll leave it up to you whether you want to resolve them or not. It looks like tests aren't building though, can you fix those up first?

MattKiazyk:
Updated from feedback.

One downside with the Apple Datasource is that there is a check in `updateAvailableXcodes` that checks the beta site and if it's available, it would chose the data from beta site over /more. 

This causes the fileSize not to show as the beta site doesn't include it.

interstateone:
> One downside with the Apple Datasource is that there is a check in updateAvailableXcodes that checks the beta site and if it's available, it would chose the data from beta site over /more.
> 
> This causes the fileSize not to show as the beta site doesn't include it.

We could change that to "merge" the two AvailableXcodes, taking the Version from `/downloads` and the file size from `/downloads/more`, instead of only taking the data from `/downloads`.

PR #100, opened 2021-02-05
Improve how identical builds are represented in the UI

### LINKED ISSUE #90: Identical builds aren't represented clearly
Hi!
I installed the app and ran it and noticed it was out of date as it didnt include 12.4 as my command line did
Screen Shot 2021-01-28 at 8 46 56 AM
Screen Shot 2021-01-28 at 8 48 56 AM
I hit the refresh button and it did not show up.  Then I went to preferences and signed in, hit refresh, still nothing. Not sure why its missing the 12.4 release (not the RC)

interstateone:
This change attempts to improve how different versions with the exact same build identifier are represented in the UI.

Background: Sometimes the last prerelease version (for example 12.4 Release Candidate) and the release version (12.4) are the exact same and have the same build identifier as a result (12D4e in this case.) While we _could_ just do the easy thing and have independent rows in the list for each version, this makes it difficult to determine which versions are installed. I also doubt that anyone would really want more than one version installed when they're the exact same build. If anything, most users would want to know when they're the exact same so that they can _avoid_ installing another version.

Through version 1.0.3 (4) I'd attempted to handle this by filtering out prerelease versions that had the same build identifier as a release version. For a couple reasons this ended up being confusing, and resulted in issue #90.

The main change here is instead of filtering out versions, keep the release version and associate the other prerelease versions with identical build identifiers with it. These identical builds can then be represented in the list and info pane.

One of the issues before was that there is some "adjusting" logic that tries to improve the data we have about available and installed Xcode versions. This was originally added because of shortcomings of scraping the Apple Developer site. I kept this code even after we added the Xcode Releases data source, and while it was mostly benign it made this identical build ambiguity worse. This PR changes this "adjusting" logic to only apply to the Apple data source instead. It also now treats what Xcode Release identifies as GM versions (which are final releases, as opposed to GM seeds) as release versions without any prerelease identifiers.

The UI changes that I ended up with are an icon in the list row and a new section in the info pane. I had tested out an expanding row that revealed the identical build versions, which I think is valuable, but I'm finding too many issues with the macOS List implementation (performance and memory leaks) and I'm concerned about making the row views more complex. I'd like to revisit this design in the future, though. Instead a single icon is added with some help text. For both of the new views I made sure to use some accessibility modifiers so they both read reasonably well with VoiceOver when testing with the Accessibility Inspector. I didn't go any further than the new UI, but I think we should take a pass on the rest of the app soon.

Screen Shot 2021-02-04 at 9 16 28 PM

_Example: note that I have 12.4 RC installed, and the row for the final release is appearing as installed with the correct path. The row has the identical builds indicator, and both versions are listed under the new heading in the info pane._

---
Closes #90

### DIFF
--- a/Xcodes/Backend/AppState+Update.swift
+++ b/Xcodes/Backend/AppState+Update.swift
@@ -6,10 +6,6 @@ import SwiftSoup
 import struct XCModel.Xcode
 
 extension AppState {
-    private var dataSource: DataSource {
-        Current.defaults.string(forKey: "dataSource").flatMap(DataSource.init(rawValue:)) ?? .default
-    }
-    
     func updateIfNeeded() {
         guard
             let lastUpdated = Current.defaults.date(forKey: "lastUpdated"),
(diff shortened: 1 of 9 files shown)


interstateone:
One thing that might be better is to use the same "rectangle.stack.fill" image in the info pane, instead of the question mark, so that there's a visual link between the (unexplained, other than help text) icon in the list row and the section in the info pane. Let me know what you think.

Screen Shot 2021-02-05 at 5 50 57 PM

MattKiazyk (changes requested):
If you use the Apple data source, it gets a little confused. 

image

interstateone:
> If you use the Apple data source, it gets a little confused.

😬 Good catch, I think I've fixed that. It also revealed that the InfoPane content should be in a ScrollView, and now it is.

|Xcode Releases|Apple|
|-----|-----|
|Screen Shot 2021-02-06 at 3 30 11 PM|Screen Shot 2021-02-06 at 3 30 22 PM|

I also settled on the "square.fill.on.square.fill" image in both spots. Ideally it would probably be an "app.fill.on.app.fill" image but that doesn't exist and after spending a bit of time on it seems way too tedious to do correctly.

MattKiazyk (approved):
👍

paulthorsteinson:
Thanks @interstateone & @MattKiazyk !!! :)

babbage:
I am a little bit unclear what the final intended result was of these changes, so am not sure whether what I am observing is a regression, or whether I am requesting consideration of a feature change.

At present, using the Apple data source, there are separate entries in the list for both "14.2 Release Candidate" and "14.2 (14C18)", even though the Release Candidate was already build 14C18, and it became the final release. In my view, the release candidate that actually becomes the final release should probably not be shown in the list, or if it is, there should be a specific indication that it is the same build as the final release. (None of the non-installed versions show a build number. Perhaps this is a limitation in the Apple data source?)

However, the behavior that I have observed and wanted to highlight is that when a Release Candidate becomes the final release, Xcodes notifies me that a new version of Xcode is available, yet when I visit Xcodes it simply shows me that I already have the final release installed, without further explanation as to what triggered the notification.

I speculate that the behavior I observe may be associated with the fact that I have the setting "Always rename to Xcode.app" turned on. (I love this setting, it keeps my dock entries working.)

If this would be better posted as a separate issue, I'm happy to do so. However, it seems (potentially) directly related to the original issue here so I thought I'd post here first.

babbage:
Changing my data source to Xcode Releases removes the Release Candidate entry, as I was suggesting would be good, and shows build numbers for historical installs. I do not know what effect this would have on notifying me if a RC build becomes an actual release.

It does make me wonder about a data source option that combines the best of both words: using Xcode Releases for all historical data, while also presenting any newer builds that are present on the Apple website but not yet incorporated into the Xcode Releases dataset?

PR #102, opened 2021-02-05
When downloading, use the downloader from userdefaults

### LINKED ISSUE #101: Download will always use aria2. It should use the preference
**Describe the bug**
In `AppState` when we go to download, the install is always set to aria - 

`self.install(.version(availableXcode), downloader: .aria2)`

We should switch that to the AppStorage on the preferences

MattKiazyk:
Fixes #101 

The install command was hardcoded to .aria2, not using the preferences option that is saved in userDefaults. 

Chose urlSession if there is no key.

### DIFF
--- a/Xcodes/Backend/AppState.swift
+++ b/Xcodes/Backend/AppState.swift
@@ -49,7 +49,7 @@ class AppState: ObservableObject {
 
     @Published var error: Error?
     @Published var authError: Error?
-    
+
     // MARK: - Publisher Cancellables
     
     var cancellables = Set<AnyCancellable>()
@@ -299,7 +299,7 @@ class AppState: ObservableObject {
                     .mapError { $0 as Error }
             }
             .flatMap { [unowned self] in
-                self.install(.version(availableXcode), downloader: .aria2)
+                self.install(.version(availableXcode), downloader: Downloader(rawValue: UserDefaults.standard.string(forKey: "downloader") ?? "urlSession") ?? .urlSession)
             }
             .receive(on: DispatchQueue.main)
             .sink(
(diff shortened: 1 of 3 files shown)


Xcodes/Resources/Licenses.rtf
```diff
@@ -1,4 +1,4 @@
-{\rtf1\ansi\ansicpg1252\cocoartf2577
+{\rtf1\ansi\ansicpg1252\cocoartf2578
```
interstateone, line 1: Could you remove this change from the branch? Unless this 2578 is a new version number or something that's going to keep appearing in diffs...

Xcodes/Backend/AppState.swift
```diff
@@ -299,7 +299,7 @@ class AppState: ObservableObject {
                     .mapError { $0 as Error }
             }
             .flatMap { [unowned self] in
-                self.install(.version(availableXcode), downloader: .aria2)
+                self.install(.version(availableXcode), downloader: Downloader(rawValue: UserDefaults.standard.string(forKey: "downloader") ?? "urlSession") ?? .urlSession)
```
interstateone, line 302: I think we should fall back to .aria2 to match the existing behaviour and the code in AdvancedPreferencePane.

```suggestion
                self.install(.version(availableXcode), downloader: Downloader(rawValue: UserDefaults.standard.string(forKey: "downloader") ?? "aria2") ?? .aria2)
```

interstateone (approved):
Works well, thanks Matt! Lets default to .aria2, but otherwise merge at will 🚢

Xcodes/Resources/Licenses.rtf
MattKiazyk: 🤷  I'll change it back

PR #104, opened 2021-02-07
Moves the .help for the ProgressIndicator so it refreshes with the fraction change

### LINKED ISSUE #93: Download step progress indicator tooltip doesn't update
**Describe the bug**

Download step progress indicator tooltip doesn't update

**To Reproduce**

1. Install a version
2. Wait a bit until the download progress indicator is showing a reasonable amount of progress
3. Hover your mouse over the progress indicator so the tooltip is presented, and see that the percentage shown doesn't match the progress indicator

**Expected behavior**

The tooltip should reflect the same completion as the progress indicator

**Screenshots**

Screen Shot 2021-01-29 at 8 58 26 AM

**Version**
 - OS: 11.1
 - Xcodes: 1.0.3

MattKiazyk:
Fixes #93 

Move the `.help` away from the StepView into the ProgressIndicator so that it can catch the fraction completed events.

### DIFF
--- a/Xcodes/Frontend/Common/ObservingProgressIndicator.swift
+++ b/Xcodes/Frontend/Common/ObservingProgressIndicator.swift
@@ -42,6 +42,7 @@ public struct ObservingProgressIndicator: View {
             isIndeterminate: progress.progress.isIndeterminate,
             style: style
         )
+        .help("Dowloading: \(Int((progress.progress.fractionCompleted * 100)))% complete")
     }
 }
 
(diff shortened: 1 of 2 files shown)


Xcodes/Frontend/Common/ObservingProgressIndicator.swift
```diff
@@ -42,6 +42,7 @@ public struct ObservingProgressIndicator: View {
             isIndeterminate: progress.progress.isIndeterminate,
             style: style
         )
+        .help("Dowloading: \(Int((progress.progress.fractionCompleted * 100)))% complete")
```
senmu, line 45: Should we fix the typo here?
```suggestion
        .help("Downloading: \(Int((progress.progress.fractionCompleted * 100)))% complete")
```

senmu (approved):
With these changes, I'm finding that the percentage reported in the "tooltip" seems to reflect what is being visualized.

Just noticed a typo that we might as well fix since that code is being touched.

senmu (approved):

PR #107, opened 2021-02-14
Display installation steps with download stats in info pane

### LINKED ISSUE #97: Add progress to sidebar view as well
When downloading it can tike quite a long time without any visible progress since the indicator is tiny. I would appreciate a bigger indicator (maybe a linear progress bar) being added to the sidebar, maybe even with a time remaining estimate for the step and a display for the current step.

I think the space between the cancel button and the horizontal etched line would be best for the progress bar, the step label could be next to the cancel button.

Alternatively the long progress bar could be the along the lower edge of the blue highlight in the main table.

MattKiazyk:
When using the Aria2 Downloader, we can easily get the downloads stats out from the output of aria2 including download speed, and ETA remaining. 

All of this information is now added to the info pane to show while Xcodes is downloading

Aria2:

https://user-images.githubusercontent.com/1119565/107977870-776c9080-6f81-11eb-8afb-e7963c31d92e.mp4

URLSession:
image

I renamed `InstallationStepView` to `InstallationStepRowView` to better distinguish the Progress change for the row vs the new `InstallationStepDetailView`

Also for the progress, I've repurposed the `totalUnitCount`. So instead of using 1-100, it uses the total bytes downloaded and the total bytes of the file. With this change, i changed the KVO for the progress to be `completedUnitCount` vs `fractionCompleted` because it would get observed more often vs the percent based. 

Inside of `downloadStats(for xcode: Xcode)` on the infoPane, my thoughts would be sometime in the future, that could change to update for the various installation steps. Currently, once it's done downloading it will go away.

___
Closes #97


interstateone (approved):
This is a nice addition Matt 👌

Xcodes/Frontend/Common/ObservingProgressIndicator.swift
```diff
@@ -26,23 +29,31 @@ public struct ObservingProgressIndicator: View {
...
+                maxValue: 1.0,
+                doubleValue: progress.progress.fractionCompleted, 
+                controlSize: controlSize,
+                isIndeterminate: progress.progress.isIndeterminate,
+                style: style
+            )
+            .help("Downloading: \(Int((progress.progress.fractionCompleted * 100)))% complete")
+            
+            if showsAdditionalDescription, progress.progress.localizedAdditionalDescription.isEmpty == false { 
+                Text(progress.progress.localizedAdditionalDescription)
```
interstateone, line 52: I changed to use localizedAdditionalDescription to let Progress do the formatting for us. This should also work with the URLSession downloader, albeit with less download info.

|aria2|URLSession|
|-----|-----|
|Screen Shot 2021-02-17 at 10 18 24 PM|Screen Shot 2021-02-18 at 7 21 50 PM|

Issue #117, opened 2021-03-29
Installing an Xcode version with an M1 may fail


MattKiazyk:
**Describe the bug**
installing an Xcode on an M1 Mac when that Mac has not yet installed Rosetta will fail, I think because aria2 isn’t a universal binary and so it requires Rosetta, but the prompt to install Rosetta isn’t shown in this case. Not sure if that’s expected or not.

“The operation couldn’t be completed. Bad CPU type in executable.”

Anybody have an M1 to test?

senmu:
It does indeed, fail
<img width="900" alt="Screen Shot 2021-04-14 at 3 15 29 PM" src="https://user-images.githubusercontent.com/818102/114780342-7f07a600-9d34-11eb-8b6e-11738bf20861.png">

MattKiazyk:
Closed via #121

Added #122 so that we can track creating of a universal binary

PR #120, opened 2021-04-15
Disable aria2 as an option for M1 macs

senmu:
One possible solution for issue #117 where only the URLSession downloader is made available.

I'm not fully convinced this is a good solution (the best thing would be a universal or arm64 binary of aria2), but this does make Xcodes.app work on M1 Macs.

### DIFF
--- a/Xcodes/Backend/AppState.swift
+++ b/Xcodes/Backend/AppState.swift
@@ -66,6 +66,15 @@ class AppState: ObservableObject {
         Current.defaults.string(forKey: "dataSource").flatMap(DataSource.init(rawValue:)) ?? .default
     }
     
+    private var preferredDownloader: Downloader {
+        #if arch(arm64)
+        // aria2 is not yet compatible on M1 silicon
+        return Downloader.urlSession
+        #elseif arch(x86_64)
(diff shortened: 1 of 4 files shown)


Xcodes/Backend/AppState.swift
```diff
@@ -66,6 +66,15 @@ class AppState: ObservableObject {
         Current.defaults.string(forKey: "dataSource").flatMap(DataSource.init(rawValue:)) ?? .default
     }
     
+    private var preferredDownloader: Downloader {
+        #if arch(arm64)
```
MattKiazyk, line 70: Does not Aria 2 work after Rosetta is installed?

senmu: I've been actively trying to avoid installing Rosetta because I'm not aware of a way to easily uninstall it so I can test the scenario without it

MattKiazyk: I'm wondering if instead of stopping aria 2, we catch the `bad cpu type` error and just show a better message?

senmu:
Closing this. I managed to compile aria2 for Apple silicon! 🎉 
Screen Shot 2021-04-15 at 5 25 37 PM

PR #121, opened 2021-04-15
Bundle M1-compatible version of aria2

senmu:
A better solution to #117.

Instead of disabling aria2 as a downloader option (attempted in #120), work was done to get another version of the binary compiled so it's compatible with m1. 
Screen Shot 2021-04-15 at 5 25 37 PM
This means the app will be inflated as it now contains two copies of aria2, but I'm unaware of how to compile a universal binary to satisfy intel and Apple's M1 in the C++ ecosystem.


MattKiazyk (approved):
Thanks @senmu - the extra 4 MB is worth it. I'll create another ticket to add a universal aria2 at some point in the future.

PR #123, opened 2021-04-23
Create universal aria2 binary

### LINKED ISSUE #122: Create a universal binary of aria2
Currently we have 2 separate binaries for aria2 to handle M1 Macs. We should try our best to have a universal binary.

senmu:
Closes #122 

## What changed
Managed to create a "fat" binary generated via: 
```bash
lipo -create -output aria2c-universal aria2c aria2c-arm64
``` 
which contains support for both x86_64 and arm64 as per Apple's documentation on [Building a Universal macOS Binary](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary#Update-the-Architecture-List-of-Custom-Makefiles). This isn't exactly an improvement in terms of reducing file size as the `aria2c` binary is now 8 MB, but it does remove the need for the conditional compilation macros.


MattKiazyk:
Thanks @senmu - eventually one day I'll get an M1 Mac!

PR #126, opened 2021-04-29
Adds MacOS notifications

### LINKED ISSUE #116: Mac OS Notifications

As a user, I would like the option to enable notification, so I can be _notified_ when certain things happen on XcodesApp

Since we now have automatic installs, would be nice to be notified when it installed itself, or when there would be a new version to install. 

image

MattKiazyk:
Closes #116 

This adds MacOS notifications to Xcodes. Basic notifications are shown for each installation step.

~~I couldn't find a way to open up the notification preferences via a url. I would like to, if the user deny's notifications, the ability to open up the MacOS notifications.~~

I think we should also somehow make the notification prompt more easy to find. Users won't go to the preferences too ofthen to turn them on. 

Preferences: 
image

Prompt:
image

Approved:
image

Denied:
image


Xcodes/Backend/AppState+Install.swift
```diff
@@ -442,6 +443,9 @@ extension AppState {
         DispatchQueue.main.async {
             guard let index = self.allXcodes.firstIndex(where: { $0.version.isEquivalent(to: version) }) else { return }
             self.allXcodes[index].installState = .installing(step)
+            
+            let xcode = self.allXcodes[index]
```
aerickson14, line 447: we're grabbing the index in the guard above, then reverting back to grabbing the instance. What about just grabbing the actual instance right from the start

```
guard let Xcode = self.allXcodes.first{where: ...
```

Xcodes/Backend/NotificationManager.swift
```diff
@@ -0,0 +1,100 @@
...
+    case finishedInstalling
+
+    public var id: Self { self }
+    
+    public var description: String {
+        switch self {
+            case .newVersionAvailable:
+                return "New version is available"
+            case .finishedInstalling:
+                return "Finished Installing"
```
aerickson14, line 27: picky but should installing use a  lowercase i?

Xcodes/Backend/Environment.swift
```diff
@@ -228,6 +229,11 @@ public struct Defaults {
     public func removeObject(forKey key: String) {
         removeObject(key)
     }
+    
+    public var get: (String) -> Any? = { UserDefaults.standard.value(forKey: $0) }
+    public func get(forKey key: String) -> Any? {
+        get(key)
+    }
```
aerickson14, line 236: are these being used anywhere? or just added for completeness

Xcodes/Frontend/Preferences/NotificationsView.swift
```diff
@@ -0,0 +1,38 @@
...
+                        .fixedSize(horizontal: false, vertical: true)
+                    Button("Notification Settings", action: {
+                        NSWorkspace.shared.open(URL(string: "x-apple.systempreferences:com.apple.preference.notifications")!)
+                    })
+                 
+                default:
+                    Button("Enable Notifications", action: {
+                        Current.notificationManager.requestAccess()
+                    }
+                    )
```
aerickson14, line 24: picky: the `}` and `)` are on different lines here but the button above they are grouped together. We should try and be consistent on formatting one way or another.

Xcodes/Backend/AppState+Install.swift
MattKiazyk: We need to set the install state of the `allXcodes` as the lists use that change of install state to change it's UI

Xcodes/Backend/Environment.swift
MattKiazyk: I think for completeness

aerickson14 (approved):

PR #127, opened 2021-04-30
Handle required log in scenarios better instead of showing errors

### LINKED ISSUE #22: Improve experience when an error requires the user to authenticate
Currently:

1. Sign out
2. Refresh
3. Note that an invalid session error alert is shown, as well as the sign in UI in preferences

Instead of showing both of these at the same time without explanation, we probably want to do something like one of these options:

- Show the error alert with a Sign In button that shows the sign in UI. There should be a way to indicate that certain errors require the user to sign in. I prefer this option.
- Skip the error alert, and only show the sign in UI
- Only show the error alert, and instruct the user to sign in in preferences.

aerickson14:
### Whats changed
- Closes #22. If login is required by an action such as tapping Install on one of the Xcode versions the user will be prompted with a login view rather than an error alert.
- Added a login button to the main toolbar

### How to test
- Launch the app starting from a state where you are not logged in 
- Tap the person icon in the main toolbar and verify you are presented with the login view
- Close the login view for now without signing in
- Tap Install on an Xcode version you currently don't have installed
- You should be prompted with a login sheet
- After signing in the install should continue as normal (An exception to this is if you get challenged by 2FA then the download may not happen automatically. Maybe we want to track this as a separate bug?)
- After successful login you can tap the person icon in the main toolbar again
- This time you should be presented with a view where you are shown as signed in
- Tapping log out from this view should dismiss the signed in view

### DIFF
--- a/Xcodes.xcodeproj/project.pbxproj
+++ b/Xcodes.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		536CFDD2263C94DE00026CE0 /* SignedInView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 536CFDD1263C94DE00026CE0 /* SignedInView.swift */; };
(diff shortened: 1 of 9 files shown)


MattKiazyk (approved):
This works a lot cleaner for when we need to show a login screen instead of the just showing an error. Thanks Andrew!

PR #129, opened 2021-05-01
Show sign in errors inline on sign in view

aerickson14:
### What's changed
- Removed the functionality where authentication errors would present an alert
- Added an inline error message to the sign in view for authentication errors
    Screen Shot 2021-05-01 at 9 14 32 AM

- Handle the case where authentication responds with an account locked error
   Screen Shot 2021-05-01 at 9 14 44 AM

### How to test
- From an unauthenticated state open up the sign in view either from preferences or the icon in the menu bar
- Fill in your username but provide an incorrect password
- You should see an inline error message "Invalid username and password combination"
- Tapping cancel or next should clear this error message
- If you input a username that is common/short/not an email address like "test" the app will now show a proper error message for account locked. Previously this would generate an alert of the form "Received an unexpected sign in response."


MattKiazyk (approved):
🎉 
So much cleaner - thanks!

PR #130, opened 2021-05-01
include updates to sparkle license

aerickson14:
AcknowledgementsGenerator has detected updates to the Sparkle license. These changes are automatically applied using this tool.


MattKiazyk (approved):
Not sure why these show up now, as they should've always been in there, but I got it too on my laptop

We use this to show the libraries we use on the acknowlement screen

PR #131, opened 2021-05-01
only present a single alert at time

### LINKED ISSUE #128: Duplicate alerts
**Describe the bug**
Actions that result in errors that show alerts will prompt the same alert multiple times

**To Reproduce**
- Perform an action that results in an error. Easiest one to reproduce is signing in with invalid credentials.
- When the alert is shown on screen wait a while (The longer you wait, the more repeated alerts you seem to get)

**Expected behavior**
Only one alert per error should ever occur

**Demo Video**
https://user-images.githubusercontent.com/1420670/116783635-03546b80-aa4d-11eb-9647-8f764e5d46a0.mov

**Version**
 - OS: 11.2.3
 - Xcodes: 1.0.5 (6)

aerickson14:
Closes #128 

### Whats changed
Refactored how alerts are presented so its not possible to present multiple alerts at once. The previous implementation using overlays was causing an alert to be displayed anytime the parent view got refreshed.

### How to test
One interesting way to test this is to change the Run configuration from Debug to Test (in Edit Scheme). This will cause an issue where privelidged helper cannot install.
- Running a build using Test you can then try and switch Xcode versions either with the checkmark icon in the list, or using "Make active" button in the info panel when an installed version of Xcode is selected.
- After tapping Install and providing account credentials to the helper you will see an error alert 
    Screen Shot 2021-05-01 at 12 56 57 PM
- Dismiss this alert and you should no longer see a duplicate error alert pop up

### DIFF
--- a/Xcodes.xcodeproj/project.pbxproj
+++ b/Xcodes.xcodeproj/project.pbxproj
@@ -9,6 +9,7 @@
 /* Begin PBXBuildFile section */
 		536CFDD2263C94DE00026CE0 /* SignedInView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 536CFDD1263C94DE00026CE0 /* SignedInView.swift */; };
 		536CFDD4263C9A8000026CE0 /* XcodesSheet.swift in Sources */ = {isa = PBXBuildFile; fileRef = 536CFDD3263C9A8000026CE0 /* XcodesSheet.swift */; };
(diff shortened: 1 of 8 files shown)


MattKiazyk:
This is cleaner for sure @aerickson14 thanks.

One duplicate I still get is:
1. Switch your datasource to Apple
2. Sign out of the apple account
3. Hit refresh
4. Tap cancel on the sign in view
5. I see 2 alerts here.

aerickson14:
a7e36ca should suppress the alert when there is an invalid session and the user refreshes the list

MattKiazyk (approved):

Issue #151, opened 2021-09-15
Support Beta version filter


Binlogo:
**Tell us how we can improve Xcodes**

- Filter button is support to switch to installed Xcode versions, it's convenient. If it support filter Beta version, it would be a nice feature.

**Is your feature request related to a problem? Please describe.**

No.

**What would you like to see? How would you like it to work?**

bdunay3:
+1 to this enhancement request

PR #153, opened 2021-09-15
Add new filter for Release or Beta versions

Binlogo:
Implement #151 

Screenshots:

Xcodes filter - all

Xcodes filter - release only

Xcodes filter - beta only

### DIFF
--- a/Xcodes/Frontend/XcodeList/XcodeListView.swift
+++ b/Xcodes/Frontend/XcodeList/XcodeListView.swift
@@ -32,7 +32,7 @@ struct XcodeListView: View {
         }
         
         if isInstalledOnly {
-            xcodes = appState.allXcodes.filter { $0.installState.installed }
+            xcodes = xcodes.filter { $0.installState.installed }
         }
         
         return xcodes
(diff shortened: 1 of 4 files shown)


Xcodes/Frontend/XcodeList/XcodeListView.swift
```diff
@@ -7,26 +7,34 @@ struct XcodeListView: View {
...
+        case .beta:
+            xcodes = appState.allXcodes.filter { $0.version.isPrerelease }
         }
         
         if !searchText.isEmpty {
             xcodes = xcodes.filter { $0.description.contains(searchText) }
         }
         
+        if isInstalledOnly {
+            xcodes = appState.allXcodes.filter { $0.installState.installed }
```
MattKiazyk, line 35:
```suggestion
            xcodes = xcodes.filter { $0.installState.installed }
```
I think it would be better if the `isInstalledOnly` is on, we filter with `AND` instead of `OR`, so that the user can use both filters.

MattKiazyk (changes requested):
Thanks @Binlogo! An awesome addition to Xcodes!

One quick change I'm going to request is making the filters be an `AND` instead of an `OR` so that there's no confusion.

Xcodes/Frontend/XcodeList/XcodeListView.swift
Binlogo: Good to go, this is my purpose too, I missed it.

MattKiazyk (approved):

Issue #190, opened 2022-03-11
Unable to install Xcode


jordanekay:
**Describe the bug**
I am unable to install Xcode using Xcodes.

**To Reproduce**
Launch app, log in, choose an Xcode version to install.

**Expected behavior**
The version should install.

**Screenshots**
I see an error message.

<img width="915" alt="image" src="https://user-images.githubusercontent.com/191970/157985247-6953cbfb-cb4d-46e0-8e84-5f255a7c1604.png">

**Version**
 - OS: macOS 12.3 beta
 - Xcodes: 13.0 (10)

senpng:
login first

jordanekay:
I am logged in.

<img width="652" alt="image" src="https://user-images.githubusercontent.com/191970/158406891-bb828808-3b31-449e-9347-207e8cb2ccfc.png">

MattKiazyk:
Hi @jordanekay 

I installed 12.3 tonight and don't get any issues on my machine. Which xcode version are you trying to install? Do you happen to have a different language or region set, or a different language somewhere set on Apple?

It's trying to parse a json that's from developer.apple.com and it's returning a different format for some reason. If you have a proxy to watch the url's it would be awesome to see what the difference is on the `listDownloads.action` url

jordanekay:
I’m trying to install the 13.3 release candidate. I have no different language or region set.

Looking at the JSON in the response body via Proxyman, there are 2,240 `dateModified` fields and they all seem to have the correct date format.

[downloads.zip](https://github.com/RobotsAndPencils/XcodesApp/files/8296709/downloads.zip)

jordanekay:
Yes, all versions have the same issue. No version can be installed through Xcodes.app.

artemsmikh:
I've got the same issue. I can help with debugging the error if needed.

SimonRice:
I was also unable to install the latest version of Xcode via Xcodes with my system settings, with exactly the same error as above - I was attempting to install Xcode 12.3 on a machine that hadn't yet had that version installed.  As an aside, as a Brit, I had my locale set to en-GB - however, unusually I set my time format set to 12h.  The screenshot below shows my "Language & Region" settings:

<img width="780" alt="Screenshot 2022-04-09 at 12 29 57 pm" src="https://user-images.githubusercontent.com/548577/162571269-a85741f0-6d9f-4133-a6ba-9d41f17763b0.png">

[...]

PR #196, opened 2022-04-09
Use a fixed locale for API response date parsing

aerickson14:
Fixes: #190 

### What was the issue
As discussed in #190, there seem to be some regions and combinations of 24-hour time preference that cause return unexpected results for the date formatter used when parsing API responses.

With an region such as `en-GB` selected on a machine under System Preferences > Language and Region, and 24-hour unchecked, when using a date formatter with the string `"MM/dd/yy HH:mm"` it will actually return a string that includes the am/pm

24hour-disabled

### What's changed
Use a fixed locale of `"en_US_POSIX"` for date parsing of api responses.

### How to test

- Go to System Preferences > Language and Region on your machine
- Select Region: Europe > United Kingdom
- Uncheck `24-Hour Time`
- Launch the Xcodes app
- Try and download any Xcode version that you do not currently have installed on your machine
- Another thing to test is refreshing the list of Xcodes using the refresh button in the top menu of the app

### DIFF
--- a/Xcodes/Backend/DateFormatter+.swift
+++ b/Xcodes/Backend/DateFormatter+.swift
@@ -5,6 +5,7 @@ extension DateFormatter {
     static let downloadsDateModified: DateFormatter = {
         let formatter = DateFormatter()
         formatter.dateFormat = "MM/dd/yy HH:mm"
+        formatter.locale = Locale(identifier: "en_US_POSIX")
         return formatter
     }()
 


MattKiazyk (approved):
Thanks @aerickson14

bartzy:
@MattKiazyk What is usually the release cycle for Xcodes.app?
Thanks!

MattKiazyk:
@bartzy no definitive set release cycle. I was thinking I would finish up the symlink additions + this and make a release.

Don't have a specific timeline for it though.

PR #200, opened 2022-04-14
Add Preference to make Symbolic Link Automatic on Xcode Select

### LINKED ISSUE #180: Add option to always have a version named "Xcode.app" on activate
This app is amazing and I use it all the time. The only thing I need to do manually right now is after upgrading to a new stable release and making it the active one, to rename it so it does not have the version in its name, which I don't want to have for the latest stable release.

**Is your feature request related to a problem? Please describe.**
I know that "make active" is the right approach, but some tools expect to have a path just called `Xcode.app` and therefore it's safer to have the latest stable just called that. See https://github.com/Flinesoft/BartyCrouch/issues/179 for example.

Besides the existing "Active" button, there could also be a "Default name" button. Or alternatively, there could be an option in the preferences stating something like "Remove version suffix from currently active Xcode apps name".

MattKiazyk:
Adds an option to Advance Preferences to automatically update the Symbolic Link for Xcodes.app to whatever Xcode version you've selected. 

image

Closes #180

### DIFF
--- a/Xcodes/Backend/AppState.swift
+++ b/Xcodes/Backend/AppState.swift
@@ -68,7 +68,13 @@ class AppState: ObservableObject {
             Current.defaults.set(unxipExperiment, forKey: "unxipExperiment")
         }
     }
-
+    
+    @Published var createSymLinkOnSelect = false {
+        didSet {
+            Current.defaults.set(createSymLinkOnSelect, forKey: "createSymLinkOnSelect")
+        }
+    }
+    
     // MARK: - Publisher Cancellables
     
     var cancellables = Set<AnyCancellable>()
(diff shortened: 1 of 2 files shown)


DD-P:
Typo: should be "making _an_ Xcode version"

Jeehut:
Thank you for this! 💯

But there's a typo in the release notes of version 1.4.0 stating Xcodes.app, I think it should be Xcode.app.

MattKiazyk:
Thanks @Jeehut typing is hard today

Jeehut:
Same typo in the tweet: https://twitter.com/xcodesapp/status/1514709515740008448?s=21&t=VTQl2DMwWvAcA1P0jQpeCg

Feel free to correct it before I retweet. 😉

MattKiazyk:
I need an edit button on Tweets 😆 

I'll just add a reply. I've updated the release notes

federicopolesello:
Yeah great it works thank you very much! Only one thing, once I restart Xcodes.app the new option preference is restored to false. Is It correct or a bug?

MattKiazyk:
@federicopolesello cheers - i see what it is - tracked in https://github.com/RobotsAndPencils/XcodesApp/issues/201

PR #212, opened 2022-04-25
Hindi Localization

KGurpreet:
## What's Changed
- Added Hindi Localization File
- Localized all strings except few words left in English for better understanding.


Xcodes/Resources/hi.lproj/Localizable.strings
```diff
@@ -0,0 +1,217 @@
...
+"MacOSRequirement" = "macOS %@ या बाद के संस्करण की आवश्यकता है";
+"SDKs" = "SDKs";
+"Compilers" = "संकलनकर्ता";
+"DownloadSize" = "डाउनलोड आकार";
+"NoXcodeSelected" = "कोई Xcode चयनित नहीं";
+
+// Installation Steps
+// When localizing. Items will be replaced in order. ie "Step 1 of 6: Downloading"
+// So if changing order, make sure the positional specficier is retained. ex: "%3$@: Step %1$d of %2$d"
+"InstallationStepDescription" = "%d: %@ का चरण %d";
```
MattKiazyk, line 44: @KGurpreet Can we update this localization to follow the comment from above? Ie put in the positional markers so that it doesn't crash trying to replace a string with an int.

If you switch to Hindi and download a version, with the current localization it will crash.

MattKiazyk (changes requested):
Hey @KGurpreet Thanks for this translation! Greatly appreciated!

There is one change requested so that it doesn't crash when downloading a version. If it doesn't make sense let me know.

Xcodes/Resources/hi.lproj/Localizable.strings
KGurpreet: Updated PR with the required changes. Please verify.

MattKiazyk (approved):
Thanks @KGurpreet Appreciated!

PR #225, opened 2022-05-17
Add support for pasting the entire 2FA code

alexisakers:
### Goals
This PR adds support for pasting the entire 2FA code if the new string equals the length of the expected 2FA code. Currently, when pasting the 2FA code, only the last digit gets pasted.

### Notes
I think it makes sense to have this behavior only if the user didn't enter any digit so that we can avoid any weird edge cases that could override their input, but let me know if you disagree.

### Demo

https://user-images.githubusercontent.com/16192914/168834317-5358b584-261c-4f7f-93f3-0ededd350410.mov

### DIFF
--- a/Xcodes/Frontend/SignIn/PinCodeTextView.swift
+++ b/Xcodes/Frontend/SignIn/PinCodeTextView.swift
@@ -139,7 +139,15 @@ class PinCodeTextView: NSControl, NSTextFieldDelegate {
         else { return } 
         
         let newFieldText = field.stringValue
-        
+
+        // If we are focused on the first field and pasting a 6-digit string,
+        // treat it as pasting the entire code and focus on the last field
+        if code.isEmpty && newFieldText.count == numberOfDigits {
+            code = Array(newFieldText)
+            window?.makeFirstResponder(characterViews.last)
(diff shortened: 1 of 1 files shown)


Xcodes/Frontend/SignIn/PinCodeTextView.swift
```diff
@@ -139,7 +139,15 @@ class PinCodeTextView: NSControl, NSTextFieldDelegate {
         else { return } 
         
         let newFieldText = field.stringValue
-        
+
+        // If we are focused on the first field and pasting a 6-digit string,
+        // treat it as pasting the entire code and focus on the last field
+        if code.isEmpty && newFieldText.count == numberOfDigits {
```
MattKiazyk, line 145: I think we can remove the `isEmpty` check and just add a check to make sure you are on the first character field instead. That way, as long as you're on the first field, you can paste the 6 digits.

MattKiazyk (changes requested):
Thanks @alexisakers!

This really shows that MacOS needs a way to copy/paste the Trusted Device 2FA code!

Looks and works great - just one small #picky change.

PR #222, opened 2022-05-14
Adding Chinese-Simplified localizations.

megabitsenmzq:
Full translation for Chinese-Simplified.
Also, fixed a small problem in other strings.

PR #228, opened 2022-05-19
Adding Chinese-Simplified localizations

hdcola:
This is a translation made possible by a teacher and his high school student.


MattKiazyk:
Hi @hdcola and @Sichengthebest 

Thank you very much for spending the time doing this 🙇 

There is another PR #222 that was created before this one, so I'm going to work off of that one for Chinese-Simplified. It is still open, and would appreciated any comments you have on their translations if you have some time.

PR #234, opened 2022-05-25
Update Localizable.strings for Russian language

duffpod:


alexmazlov (approved):
That sounds better. Thanks!

PR #240, opened 2022-06-06
Add Traditional Chinese localization

itszero:
Picking up where @ethanhuang13 is left. Here's the full translation for Traditional Chinese.

cc @pofat


MattKiazyk:
Thanks @ethanhuang13 and @itszero for getting this over the line!

PR #243, opened 2022-06-08
German translation added

drct:
Added de.lproj folder containing Localizable.strings with German translation. No changes to Xcode's project file. First translation and github pull request for me, so I hope everything's alright. Cheers.

PR #245, opened 2022-06-08
Add German localization

schroederm:
Added german localization. Slightly widened the AdvancedPreferencePane to prevent line breaks of single characters. Setting GroupBox labels in that pane to .trailing alignment for a cleaner look.


MattKiazyk:
Hi @schroederm - Thanks for taking the time to making this PR! It is greatly appreciated!

PR #243 was made just a bit before you, and with the same translation, I'm going to take that one and close this one. 

Thanks again!

PR #258, opened 2022-06-25
Fix macOS Ventura crash on Settings View 

### LINKED ISSUE #242: Crash when navigating to the Experiments tab on macOS 13 developer beta
**Describe the bug**
When I navigate to the Experiments tab on macOS 13, the app crashes, however, when opening the app again and opening settings, the experiments tab is shown normally. Also, after navigating to another tab, the updates tab will crash the app as well. 

**To Reproduce**
Open Xcodes.app, open preferences, click on experiments.

**Expected behavior**
The app does not crash

**Version**
 - OS: macOS 13.0
 - Xcodes: 1.6.0

To be honest, I don't know if I should report this to you all or to apple, so let me know if it is more appropriate to submit a feedback to apple about this.

GetToSet:
Fixes #242.

This PR fixes a SwiftUI related crash on macOS Ventura (both beta1 and 2) when users enter the `Advanced` settings tab and then navigate to another tab.

This crash can be solved by moving pane `.frame(width:)` to its parents. I know it's tricky, but it should solve this issue.

## Test Checklist
- [x] macOS Ventura Beta2 (22A5286j)
- [x] macOS Monterey 12.4 (21F79)
- [x] macOS Big Sur 11.6.7 (20G630)

### DIFF
--- a/Xcodes/Frontend/Preferences/AdvancedPreferencePane.swift
+++ b/Xcodes/Frontend/Preferences/AdvancedPreferencePane.swift
@@ -118,7 +118,6 @@ struct AdvancedPreferencePane: View {
             }
             .groupBoxStyle(PreferencesGroupBoxStyle())
         }
-        .frame(width: 500)
     }
     
     private var dataSourceFootnote: NSAttributedString {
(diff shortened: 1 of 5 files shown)


MattKiazyk:
Thanks @GetToSet - greatly appreciated!

PR #264, opened 2022-07-14
Use monospaced digit font for download progress

gredman:
Hi,

This is a really minor change to use monospaced digits for the download progress (X.X GB of Y.Y GB), which prevents the label from jumping around so much when the numbers change.

Thanks!


MattKiazyk:
TIL about `monospacedDigit` thanks @gredman great add!

PR #267, opened 2022-07-24
add option to rename Xcode to Xcode.app on select.

### LINKED ISSUE #180: Add option to always have a version named "Xcode.app" on activate

MattKiazyk:
Fixes #180 

This will add the option on Activate/Select to rename it to `Xcode.app`. 

Users will now get 2 options for the  Activate/Select

- Default (save as `Xcode-x.x.x.app`)
- Rename Active to `Xcode.app`. 

When the latter is chosen, the user cannot select the `Create Symbolic Link`. Xcodes will try to do its best to rename the old named `Xcodes.app` to the version number (default) format. 

image


ptrkstr:
Amazing work! Thank you @MattKiazyk ❤️

PR #270, opened 2022-08-09
Update zh_Hans localizations.

megabitsenmzq:
Update zh_Hans localization file with new strings.


MattKiazyk:
Thanks @megabitsenmzq !

PR #271, opened 2022-08-10
Update Localizable.strings for Russian language (2)

duffpod:
Fixed typos here and there, made casing more consistent, rewrote some sentences to make more sense, localized new strings from en.lproj.


MattKiazyk:
Thanks @duffpod

PR #274, opened 2022-08-17
Update readme with Taiwan flag

ymhuang0808:
Hi @MattKiazyk , 

Thanks for adding localization contributors section recently. In Traditional Chinese localization, the flag is wrong. So, I change the flag into 🇹🇼 for Traditional Chinese.


ymhuang0808:
Thank you @MattKiazyk !

PR #290, opened 2022-09-17
Add ability to download Xcode without logging in using XcodeReleases

MattKiazyk:
I hate having to log into my Apple Developer account, always having to put in my 2fa code, when I want to download Xcode, but Apple has never given us the ability to do so.

This PR changes that.

By using a public URL provided by Apple, we can get a token cookie that can be used to download the version of Xcode that we wish, without having to provide a log in. 🎉 

Caveats:
- This does not work when using the `Apple` datasource. More specifically that token doesn't work for listing the Xcode versions with `https://developer.apple.com/services-account/QH65B2/downloadws/listDownloads.action`

### DIFF
--- a/Xcodes/Backend/AppState+Install.swift
+++ b/Xcodes/Backend/AppState+Install.swift
@@ -40,15 +40,10 @@ extension AppState {
     }
     
     private func install(_ installationType: InstallationType, downloader: Downloader, attemptNumber: Int) -> AnyPublisher<InstalledXcode, Error> {
-        // We need to check if the Apple ID that is logged in is an Apple Developer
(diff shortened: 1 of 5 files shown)


timsutton:
I just noticed this as well, when looking into how Xcode 14 is able to download the new DMG-based simulators that now sit behind ADC authentication, without needing a login. Super handy that it works for all downloads including Xcode! 🥳

lpusok:
@MattKiazyk trying to reproduce this feature in Postman. i successfully get ADCDownloadAuth cookie from developerservices2.apple.com, but the URL (https://developer.apple.com/services-account/download?path=/Developer_Tools/Xcode_14.0.1/Xcode_14.0.1.xip) redirects to signin page still. Any idea how to get this to work?

PR #294, opened 2022-09-20
Custom Searchbar that includes a while placeholder in dark mode

### LINKED ISSUE #293: Improving TextField foregroundColor on Dark Mode
**Describe the bug**
The placeholder color in the textfield is black on dark mode which is not good on UI

**To Reproduce**
Opening the app in dark mode

**Expected behavior**
Placeholder color should be white.

**Screenshots**
image

**Version**
 - OS: macOS Ventura Beta 8
 - Xcodes: 1.8.0

egrimo:
A workaround for close #293

### DIFF
--- a/Xcodes/Frontend/XcodeList/MainToolbar.swift
+++ b/Xcodes/Frontend/XcodeList/MainToolbar.swift
@@ -91,10 +91,16 @@ struct MainToolbarModifier: ViewModifier {
             })
             .help("PreferencesDescription")
             
-            TextField("Search", text: $searchText)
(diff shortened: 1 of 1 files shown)


marcusziade (commented):
You can add `close #<issue number>` in the PR description. If this gets merged, it will close the issue automatically.

Also, the title isn't describing what changes when this is applied.

MattKiazyk:
@egrimo i'm going to wait until the final Ventura release is out before merging, Just in case this bug is fixed with a newer version.

egrimo:
The Problem seems not be present on the Ventura RC version and the placeholder color is fixed.

PR #304, opened 2022-10-05
Adds an option to save beta versions with `-beta` symlink

### LINKED ISSUE #281: Add Xcode Beta Symlinking/Renaming Options
When working with Xcode betas, it would be great to have 2 new things (which are related I think):

1. The ability to rename the betas to `Xcode-beta.app` just like they are when downloaded from the dev portal.2. 
2. The ability to symlink an Xcode beta as `Xcode-beta.app`.3. 

The motivation for this would be to aid in tooling which can target either Xcode production (which should still be `Xcode.app`) and Xcode betas (for things like [this Xcode beta simulator fix](https://blog.nihongo-app.com/stop-xcode-14-beta-from-pegging-cpu/)).

I think this could be pulled off a couple of ways: have a new section in the preferences related to Xcode betas specifically, or add an option in the Active/Select section of the advanced preferences to symlink to the Xcode betas. In addition, right clicking on a beta would display an additional menu item to create a symlink for a beta version when one is selected. I don't think getting rid of the `Xcode.app` symlink is right since some people may still want to do that to test their workflows with a beta version.

leon-wolf:
resolves #281 

- adding .idea to .gitignore for convenience
- adding option to create Xcode-Beta.app for prerelease Items

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,9 @@ Product/*
 timeline.xctimeline
 playground.xcworkspace
 
+# Jetbrains IDEA
+.idea
+
 # Swift Package Manager
 #
 # Add this line if you want to avoid checking in source code from Swift Package Manager dependencies.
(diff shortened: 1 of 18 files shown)


pvinis (commented):
very nice

pvinis (approved):

PR #308, opened 2022-10-07
Remove localization from apple.com links

jfversluis:
While providing the Dutch translation I noticed that the link to the apple.com website had the localization identifier in it. Some people adapted it for the right language, others didn't.

But just removing it, the Apple website will automatically route a visitor to the right language based on their machine settings. I personally think that is always the best option.

Just put it in a PR if you think so too, if not, feel free to disregard :)


MattKiazyk (approved):
👍 Thanks @jfversluis - good change

PR #309, opened 2022-10-13
Fix localization on UnsupportedFileFormat

tt37:


MattKiazyk (approved):
Thanks @tt37 - nice find!

Issue #230, opened 2022-05-19
App crashes when scrolling while installing


EthanLipnik:
**Describe the bug**
The app crashes when you scroll as a version of Xcode is installing.

**To Reproduce**

- Click install on an Xcode version
- Scroll on the main scroll view

**Expected behavior**
Able to scroll down while installing.

**Version**
 - OS: 12.4
 - Xcodes: 1.5.0

MattKiazyk:
Hi @EthanLipnik 

I couldn't duplicate the crash from the steps above. See the video below.

I am on 12.3.1, so I'll try again once I update. 

https://user-images.githubusercontent.com/1119565/169447116-7450e217-2b3b-4422-aeab-d68f96bafd85.mov

DD-P:
I couldn't reproduce either.  iMac Pro, macOS 12.4, Xcodes 1.5
Filtered to release-only, info panel displayed.  Xcode 13.3.1 selected and download initiated from info panel install button.

EthanLipnik:
> Hi @EthanLipnik
> 
> I couldn't duplicate the crash from the steps above. See the video below.
> 
> I am on 12.3.1, so I'll try again once I update.
> 
>  May-19-2022.22-52-45.mov

Hi @MattKiazyk, I updated the issue with better reproduce steps and a screen recording

MattKiazyk:
Looks like you aren't even scrolling on that recording? 

If you have time, could you clone the repo and try running the app in debug on your machine?

DD-P:
At 6 secs the scrollbar thumb becomes visible, which suggests scrolling is attempted but doesn't actually occur on-screen.

uyloal:
Hi @MattKiazyk , I also met this crash. 

Feels like the crash will happen in the Xcode install steps, I met the crash when I scroll and unzip the xip file

Here is a crash reports, may be informative:

```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               Xcodes [26133]
Path:                  /Applications/Xcodes.app/Contents/MacOS/Xcodes
Identifier:            com.robotsandpencils.XcodesApp
Version:               1.8.0 (16)
Code Type:             X86-64 (Native)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2022-10-19 14:35:58.9474 +0800
OS Version:            macOS 12.6 (21G115)
Report Version:        12
Bridge OS Version:     6.6 (19P6067)
[...]

PR #311, opened 2022-10-19
Fix the ProgressView crash with List.

uyloal:
Fix issue #230 

On macOS, when a list item containing a ProgressView is selected, the accessibility of the ProgressView is treated as if it exists by default. Even if we never set the accessibility property.

I only fix `ProgressView` in `XcodeListView`. 
Maybe the other `ProgressView` need to be fixed as well and you should check them.

If you want to reproduce, here is a minimum reproducible example.
https://github.com/uyloal/progressview-crash-when-list-scroll


uyloal:
It's not a issue, caused by [AltTab](https://github.com/lwouis/alt-tab-macos)

PR #303, opened 2022-10-04
fix: Preference button on toolbar not working on Mac OS 13 Ventura

chedabob:
Apple have renamed `showPreferencesWindow` to `showSettingsWindow` which was stopping the cog icon from doing anything on Mac OS 13

PR #313, opened 2022-10-23
Fix "Open Preferences" toolbar action for macOS Ventura

ptxmac:
Apple renamed Preferences to Settings in macOS 13.0, including the selector


ptxmac:
Woops, just realized this is already done in #303 

🤦

PR #322, opened 2022-11-11
Improve Spanish localization in Preferences

DannyJJK:
I've not added all the missing ones but this fills some of them in, including some corrections


MattKiazyk:
Thanks @DannyJJK

PR #328, opened 2022-11-18
Switch Xcode releases back to using Apple Auth :(

MattKiazyk:
In Sept 2022, Xcodes was updated to use the same mechanism that is used to download runtimes with no login, to download Xcode.

ie: https://developerservices2.apple.com/services/download?path=/Developer_Tools/tvOS_16.1_Simulator_Runtime/tvOS_16.1_Simulator_Runtime.dmg returns 200 to set the ADCDownloadAuth cookie to be able to download

As of mid Nov 2022, Apple has return 403 (forbidden) for this endpoint for Xcode versions only. Runtimes still return 200.

https://developerservices2.apple.com/services/download?path=/Developer_Tools/Xcode_14.0.1/Xcode_14.0.1.xip

I'm not sure if this is deliberate from Apple (My guess is it is), but this PR reinstates the old way of download xcodes as all libraries and utilities have done for years now.

I don't think Apple will ever allow this open again. I do hope at some point, there is a way that I can use app specific passwords, or some other means, where I don't have to ask for a dev's Apple username/password.

#SadMatt

### DIFF
--- a/Xcodes/Backend/AppState+Install.swift
+++ b/Xcodes/Backend/AppState+Install.swift
@@ -43,7 +43,10 @@ extension AppState {
         
         Logger.appState.info("Using \(downloader) downloader")
         
-        return self.getXcodeArchive(installationType, downloader: downloader)
+        return validateSession()
+            .flatMap { _ in
+                self.getXcodeArchive(installationType, downloader: downloader)
+            }
             .flatMap { xcode, url -> AnyPublisher<InstalledXcode, Swift.Error> in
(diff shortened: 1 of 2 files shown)


davdroman-rdx:
There go the best 2 months of my life. Oh well.

davdroman:
There go the best 2 months of my life. Oh well.

TheMetalCode:
My dreams of unattended Xcode install for relatively pain free CI runner management, crushed to bits. 👎

PR #165, opened 2021-11-01
Show a message when user is not authorized yet

qyzdo:
Fix #164, show a message when user is not authorized instead of json decoding error.

Issue #164, opened 2021-10-27
Unable to install Xcode 13.1


vpilkevych:
**Describe the bug**
The app gives an error in attempt to install Xcode 13.1 on Mac mini M1, although it works on different Intel machines.

**To Reproduce**
1. Press the "Install" button in order to begin Xcode 13.1 installation.
2. The app shows the error message.

**Expected behavior**
The app installs Xcode 13.1.

**Screenshots**
![Screenshot 2021-10-27 at 11 08 22](https://user-images.githubusercontent.com/920844/139036265-ca268e38-24fc-4e56-9524-9c55f5374013.png)

**Version**
 - OS: 11.6, Mac mini M1
 - Xcodes: 1.2.0 (8)

qyzdo:
I wasn't logged in and I had the same error. I found that https://appstoreconnect.apple.com/olympus/v1/session is returning unauthorized status and xcodes had a problem parsing this response.

vpilkevych:
@qyzdo you're completely right, today I decided to give it one more try and found out that I was not logged in. I am sorry, my bad. Closing the issue. @MattKiazyk maybe it'd be better if the app tells the user to log in.

qyzdo:
@vpilkevych you are absolutely right, I created fix for this https://github.com/RobotsAndPencils/XcodesApp/pull/165

PR #330, opened 2022-12-03
Improve UX when the user is not logged in

PWrzesinski:
When running XcodesApp for the first time, it took me a while to figure out that I need to log in before I can download Xcode - it was a while ago, but the error was quite unclear at the time (see https://github.com/RobotsAndPencils/XcodesApp/issues/164 and https://github.com/RobotsAndPencils/XcodesApp/pull/165). This has been fixed since, but I hope making the button red when the user is not authenticated will improve the UX a little bit, guiding the user to click it.

It's a minor improvement that was at the back of my mind ever since that first time and I finally got around to implementing it. Please let me know if this kind of UI change is welcome or not.

Xcodes_red_auth_button

### DIFF
--- a/Xcodes/AppleAPI/Sources/AppleAPI/Client.swift
+++ b/Xcodes/AppleAPI/Sources/AppleAPI/Client.swift
@@ -182,6 +182,7 @@ public class Client {
 // MARK: - Types
 
 public enum AuthenticationState: Equatable {
+    case checking
     case unauthenticated
     case waitingForSecondFactor(TwoFactorOption, AuthOptionsResponse, AppleSessionData)
     case authenticated
--- a/Xcodes/Backend/AppState.swift
+++ b/Xcodes/Backend/AppState.swift
@@ -13,7 +13,8 @@ class AppState: ObservableObject {
    
     // MARK: - Published Properties
     
-    @Published var authenticationState: AuthenticationState = .unauthenticated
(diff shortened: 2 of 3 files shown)


MattKiazyk:
Hey @PWrzesinski Thanks for taking the time for this PR!

Couple of thoughts: 
- I agree that there should be something to notify the user they need to log in, I don't _mind_ the red, but I'm almost wondering if words should also appear on next to the button - similar to how filters work: 
image
- Secondly - that `authenticationState` needs some work as by default when you first open the app, it doesn't even check it, so it will always be red when you first open up the app, and only show what it actually is after it's refreshed. So I think there's some improvement there that needs to happen at the same time for the red icon on the main list to be useful. 

Let me know if you have any questions.

PWrzesinski:
Thanks for taking a look @MattKiazyk! Here's how it looks currently:

* At the start:
Xcodes_auth_spinner

* When not logged in:
Xcodes_red_auth_button_2

Considering if the red color is needed now, or perhaps the default one will do nice. I've added the check for authentication state in the AppState `init`, is that ok? Completely new to this code base.

Also got a couple of general questions:
* How do I go about localizing the new copy?
* In Contributing.md there is a mention of linting, but I can't find any configuration in the project - ok to ignore, or should I just run Swiftlint with a default config?
* Didn't add any tests for the changes, doesn't feel necessary here.

PR #336, opened 2022-12-17
Fix #329

### LINKED ISSUE #329: Error downloading any version: "Date string does not match format expected by formatter"
**Describe the bug**
I get an error downloading any version: "Date string does not match format expected by formatter"
The full message is:
```
The operation couldn’t be completed. (Error.dataCorrupted(Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "downloads", intValue: nil), _JSONKey(stringValue: "Index 2", intValue: 2), CodingKeys(stringValue: "dateModified", intValue: nil)], debugDescription: "Date string does not match format expected by formatter.", underlyingError: nil)))
```
The actual value in the JSON (at `["downloads"][2]["dateModified"]` retrieved through `mitmproxy`) is:
```
03/23/12 02:18
```

**To Reproduce**
Try downloading any version of Xcode using Xcodes

**Expected behavior**
It should work...

**Version**
 - OS: 13.0.1
 - Xcodes: 1.9.0

yossizahn:
In some time zones, date parsing of the xcodes releases json is broken since it comes across a datetime that didn't exist in that timezone due to the effects of DST.
This pull request tries fixing it as per https://stackoverflow.com/a/71415385/

### DIFF
--- a/Xcodes/Backend/DateFormatter+.swift
+++ b/Xcodes/Backend/DateFormatter+.swift
@@ -6,6 +6,7 @@ extension DateFormatter {
         let formatter = DateFormatter()
         formatter.dateFormat = "MM/dd/yy HH:mm"
         formatter.locale = Locale(identifier: "en_US_POSIX")
+        formatter.calendar = .init(identifier: .iso8601)
         return formatter
     }()
 


MattKiazyk:
Thanks @yossizahn

PR #349, opened 2023-01-16
Ventura UI fixes

FelixHerrmann:
With the update to macOS Ventura a few SwiftUI behaviors have changed. I found 2 things in the UI that worked with Monterey but are now a bit broken. I fixed both issues, but it would be good if someone could test on previous macOS versions!

| Toolbar before fix | Toolbar after fix |
| ------------------ | --------------- |
| Screenshot 2023-01-16 at 17 46 28 | Screenshot 2023-01-16 at 17 47 18 |

| Status bar before fix | Status bar after fix |
| -------------------- | ------------------- |
| Screenshot 2023-01-16 at 17 46 28 | Screenshot 2023-01-16 at 17 48 23 |

### DIFF
--- a/Xcodes/Frontend/XcodeList/BottomStatusBar.swift
+++ b/Xcodes/Frontend/XcodeList/BottomStatusBar.swift
@@ -9,27 +9,27 @@
 import Foundation
 import SwiftUI
 
-
 struct BottomStatusModifier: ViewModifier {
     @EnvironmentObject var appState: AppState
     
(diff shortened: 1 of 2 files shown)


MattKiazyk:
Thanks @FelixHerrmann for the contribution! 

I don't have Monterey but I feel that the by using the default for `ToolbarItemGroup` on Monterey it should be ok. If not hopefully somebody will let us know.

PR #361, opened 2023-02-24
Implement hashcash for Apple ID Authentication

MattKiazyk:


Xcodes/AppleAPI/Sources/AppleAPI/Hashcash.swift
```diff
@@ -0,0 +1,127 @@
...
+        } else {
+            let formatter = DateFormatter()
+            formatter.dateFormat = stampSeconds ? "yyMMddHHmmss" : "yyMMdd"
+            ts = formatter.string(from: Date())
+        }
+        
+        let challenge = "\(ver):\(bits):\(ts):\(resource):"
+        
+        var counter = 0
+        let hexDigits = Int(ceil((Double(bits) / 4)))
```
dhinakg, line 35: The number of bits isn't guaranteed to be a multiple of 4, so in cases like when bits is `10`, the code will require 3 zeros in the hex hash, however if 10 bits are zeroed then only the first two digits should be zeros + the 3rd digit should be <= 3.

dhinakg:
Not a Swift person, so this might be an abomination, but try this: 
[0001-Iterate-over-bits-of-hash-instead-of-checking-hex-di.patch](https://github.com/RobotsAndPencils/XcodesApp/files/10821462/0001-Iterate-over-bits-of-hash-instead-of-checking-hex-di.patch)

PR #383, opened 2023-05-06
Show alternating row backgrounds on XcodeListView

jmshrv:
This improves readability, expecially when doing actions like selecting the install button of a specific version. The view modifier for this requires macOS 12, so macOS 11 will not have alternating row backgrounds.

Screenshot of the app, now with alternating row backgrounds


MattKiazyk:
@jmshrv I'm not sure if I'm 100% liking it or not. Could we make an option in settings to turn off/on as a compromise?

PR #385, opened 2023-05-16
Updated readme

KyLeggiero:
- Fixed broken CI badge
- Updated screenshots for current version
- Clearer wording for installation instruction titles (now it's clear these are separate/alternative processes rather than stages in the same process)
    - Pulled brew instructions out of the codeblock so that it wraps. Rephrased for the new position


MattKiazyk:
cheers 🍻

Issue #377, opened 2023-04-18
App crashes when opening settings (cmd+,)


pietro1704:
**Describe the bug**
App crashes when opening settings. First time it happened, it was when accessing one of the settings tabs, but now, even when reinstalling the app and cleaning leftovers, still crashes when opening settings, I believe because the last tab, where the error occurs, is saved.

**To Reproduce**
Access the app settings, either via menu bar or shortcut (cmd+,)

**Expected behavior**
settings open normally and app does not crash

**Screenshots**

https://user-images.githubusercontent.com/45739756/232820583-83480e4e-6dfc-4c38-8ca7-242ae9f0014b.mov

**Version**
 - OS: MacOS Ventura 13.3.1; MacBook Pro 2017 15 in, Intel processor, and got the same error with an Macbook Pro 2018, same OS and processor
 - Xcodes: 1.10.0 (18)
<img width="488" alt="Captura de Tela 2023-04-18 às 12 10 22" src="https://user-images.githubusercontent.com/45739756/232821378-a8987674-f851-432b-bd5c-564b56ef22d4.png">

0x7067:
Also happens with me

MattKiazyk:
@pietro1704 I found the issue. In the `pt-BR` translation file under `DataSourceDescription` the app, when localizing is searching for the words `Xcode Releases` and crashes if it doesn't find it.

If you are fluent in Portuguese perhaps you can create a PR where `DataSourceDescription` value includes  `Xcode Releases` somewhere and that will fix the issue.

Workaround in the mean time is to flip Xcode not use `pt-BR` localization.

0x7067:
@MattKiazyk I fixed the issue on #386

PR #386, opened 2023-05-19
Improve Brazilian Portuguese translation

0x7067:
Hi,

This not only fixes issue #377 but also adds translation for some strings that were missing

Thanks for creating this tool!


MattKiazyk:
Thanks @0x7067

PR #398, opened 2023-06-20
Add PinCodeTextField autocomplete support

Kyle-Ye:
## Motivation

After entering last character, we need manually click the continue button.

We should follow how Apple is doing for their 2FA Service.

## Result

After entering last character, trigger the continue button automatically.

## Test

Run the Preview on PinCodeTextView.swift with Xcode 14.3

When entering last character check the "Previews log" console in Xcode.


MattKiazyk:
🍻 @Kyle-Ye (sorry for the embarrassingly long merge time)

PR #400, opened 2023-06-26
SDK column in the info panel supports displaying visionOS

LavareX:
Xcode 15 beta 2 now supports visionOS SDK.


calimarkus:
Should be good to merge it?

calimarkus:
This would resolve issue https://github.com/XcodesOrg/XcodesApp/issues/432

MattKiazyk (approved):
Thanks @LavareX 

Sorry for the late response.

Saafo:
Xcode 15.2 came with support of visionOS 1.0, should we release a new version to support this feature?

Issue #334, opened 2022-12-08
Add filter option to include releases and release candidates


klundberg:
**Tell us how we can improve Xcodes**
I usually don't want to see all xcode versions, sometimes I want to only see the released versions, which the release only filter is useful for. However, I'd love to also have the option to see release candidates in this list, since very often those RCs will just become the released version in time. 

**Is your feature request related to a problem? Please describe.**
not a problem, just a convenience i'd like added

**What would you like to see? How would you like it to work?**
Implementation isn't as important, but I could see it either being implemented as a new "Releases and RCs only" filter, this behavior just being defaulted into the "Releases Only" filter, or perhaps as a preference checkbox to "include unreleased RCs in the releases only filter". any of these (or some other solution that i haven't thought of) could work, though changing the behavior of "releases only" might not be agreeable to some folks, so making it an option would likely be the best solution for all.

jayrhynas:
I had a related but slightly different request - showing releases + betas that don't have a release. So for example right now it would show mainline Xcode releases as well as the Xcode 15 betas, but once Xcode 15 is released, those betas would hide. I could take a crack at implementing a PR but I'm also not sure the best UI for this combinatorial explosion of filters.

klundberg:
I like this option better than my proposed rc-only wording actually, and I imagine it'd be simpler to reason about and implement? Perhaps "releases + newest betas" could be the name for a new filter option (which would include RCs as well)

PR #403, opened 2023-07-07
Add `releasePlusNewBetas` XcodeListCategory

jayrhynas:
This adds a new filter option that shows only releases, plus any betas that do not have a release version. So as of today (July 7, 2023 the Xcode 15 betas will be shown but the Xcode 14 betas won't. Note that this will also show betas for older Xcode versions if necessary, for example if an Xcode 14.3.2 beta is released it will be shown until the mainline Xcode 14.3.2 is released.

This is an implementation for the ideas discussed in #334

I do find the name "Releases & New Betas" to be a little long/wordy, both in the UI and in code. I'm open to suggestions for more succinct names. This also introduces new strings that will need to be localized.


Xcodes/Frontend/XcodeList/MainToolbar.swift
```diff
@@ -58,6 +59,16 @@ struct MainToolbarModifier: ViewModifier {
                             .labelStyle(TitleOnlyLabelStyle())
                             .foregroundColor(.accentColor)
                     }
+                case .releasePlusBeta:
```
MattKiazyk, line 62: There's some errors here around the name. Looks like you may have changed around the enum name and it's invalid here.

Xcodes/Frontend/XcodeList/MainToolbar.swift
```diff
@@ -58,6 +59,16 @@ struct MainToolbarModifier: ViewModifier {
...
+                    if #available(macOS 11.3, *) {
+                        Label("ReleasePlusBetaOnly", systemImage: "line.horizontal.3.decrease.circle.fill")
```
MattKiazyk, line 64: This localized string does not exist

Xcodes/Resources/en.lproj/Localizable.strings
```diff
@@ -144,6 +144,8 @@
 "ReleaseOnly" = "Release only";
 "Beta" = "Beta";
 "BetaOnly" = "Beta only";
+"ReleasePlusNewBetas" = "Release & New Betas";
```
MattKiazyk, line 147: I think I like this one better - the `only` word is kind of redundant

MattKiazyk (changes requested):
Thanks for spending the time! I like the idea - just some clean up to make it compile and the localization string.

jayrhynas:
Thanks for looking it over! I'm not sure how I ended up committing the wrong enum names and localization keys there, sorry about that!

jayrhynas:
Fixed merge conflicts for Xcodes 2 update

jayrhynas:
Fixed merge conflicts for latest version 4.0.2

Xcodes/Resources/Localizable.xcstrings
```diff
@@ -22142,7 +22142,7 @@
         "en" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "Release & New Betas"
+            "value" : "Release & unreleased beta"
```
jayrhynas, line 22145: Still not sure the best wording for this if you have better suggestions!

PR #416, opened 2023-08-24
Enable text selection for Xcode version

### LINKED ISSUE #415: Make the version number in info pane copy-able

Can you make this text label copy-able so it is easier to get the version info? Sometimes I use the version in bug reports.

image

I can help with this improvement if you need a hand.

Thank you.

yo1995:
Close #415


Xcodes/Frontend/InfoPane/InfoPane.swift
```diff
@@ -19,6 +19,7 @@ struct InfoPane: View {
 
                     Text(verbatim: "Xcode \(xcode.description) \(xcode.version.buildMetadataIdentifiersDisplay)")
                         .font(.title)
+                        .textSelection(.enabled)
```
MattKiazyk, line 22: `.textSelection` is only available on MacOS 12. Currently Xcodes has a minimum version of 11. 

You can either wrap this in a `#available`  or this PR can wait a bit. Some point soon we'll update to min of 12.

MattKiazyk (changes requested):
Thanks! 

See comment around macOS 12 vs 11

Xcodes/Frontend/InfoPane/InfoPane.swift
yo1995: Thanks! It was an oversight from my part - I only saw the "build with macOS 12" part in README, but didn't read the "run on macOS 11.0" line 😅 

I'll defer it to your decision - feel free to wait min target becomes macOS 12 or edit this PR with

```swift
private extension InfoPane {
    @ViewBuilder
    func textSelection(_ enabled: Bool) -> some View {
        if #available(macOS 12, *), enabled {
            self
                .textSelection(.enabled)
        } else {
            self
        }
    }
}
```

MattKiazyk:
This has been added via https://github.com/XcodesOrg/XcodesApp/pull/503

PR #436, opened 2023-10-14
Clean up code in `InfoPane` to be more robust

thai-d-v:
# Problem of `InfoPane`

1. The preview code is too heavy. It crashes all the time.
2. Explicit dependencies are too broad
    - make it hard to create preview data (in reality)
    - reduce performance (in theory)
    - reduce reusability (in the future)
3. The `body` is too big
    - hard to reason
    - hard to test by preview
    
# Approach

1. I move subviews to separate files
    - reduce data dependency
    - reduce logic responsibilities, focus mostly on displaying data
    - reduce layout responsibilities. Let the outer view decide the layout
    - add preview to test locally

2. I fix `InforPane`
    - reduce data dependency
    - improve preview performance. It is fast and doesn't crash anymore
    - fix layout
    - I move `UnselectedView` out to:
        - `InfoPane` should only care about displaying data, not whether an `Xcode` is selected

# Discussion

- I want to hear from you about this direction before I can go further to improve it.

### DIFF
--- a/Xcodes.xcodeproj/project.pbxproj
+++ b/Xcodes.xcodeproj/project.pbxproj
@@ -16,6 +16,7 @@
 		B0403CF02AD92D7B00137C09 /* ReleaseNotesView.swift in Sources */ = {isa = PBXBuildFile; fileRef = B0403CEF2AD92D7B00137C09 /* ReleaseNotesView.swift */; };
 		B0403CF22AD934B600137C09 /* CompatibilityView.swift in Sources */ = {isa = PBXBuildFile; fileRef = B0403CF12AD934B600137C09 /* CompatibilityView.swift */; };
 		B0403CF42AD9381D00137C09 /* SDKsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = B0403CF32AD9381D00137C09 /* SDKsView.swift */; };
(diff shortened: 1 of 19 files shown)


MattKiazyk:
Hi @thai-d-v 

First of all thank you for taking the time to make Xcodes better!

I'm all for making `InfoPane` easier to read, easier to preview and all you've said above. This is a really good start. 

The main question I have is around all the `WrapperView`s? What is the intent of those?  I would rather see previews using the main view it's previewing vs having a wrapper. 

Thanks again - let me know how I can help

MattKiazyk (changes requested):
See comment - really awesome start - just questions around the `WrapperView`

thai-d-v:
Hi @MattKiazyk, `WapperView` provides a local interactive preview for `InfoPane`. It has some main benefits:
1. faster preview
2. local reasoning: we can preview interactively many states and layout of `InfoPane` with a small dependency

MattKiazyk:
@thai-d-v thanks for the response. I see some benefit in it, however I would rather stick with the Apple way of using different previews for each view you are previewing. Added benefit would be that the view you are working on can be previewed as you are working on. In the WrapperView scenario, I would have to change the state PreviewName to be the one I'm on every time I wanted to edit and preview. 

Feel free to use the `#Preview` macro now if you rebase from main. CI is now Xcode 15.

thai-d-v:
@MattKiazyk  You're right, code looks much cleaner with `#Preview`. I updated it.

shakeliaj:
Can you please remove me from this mailing list????

MattKiazyk (approved):

PR #437, opened 2023-10-17
Activity Indicator on Install Pressed

chickdan:
I noticed that there is often a long delay between clicking Install and the button switching to the linear progress view which sometimes makes it seem unresponsive. So I added a circular style progress view to the install button to indicate activity and reduce the chance of repeat clicks.

Before:
Screenshot 2023-10-17 at 6 21 21 PM

After:
Screenshot 2023-10-17 at 6 21 26 PM

### DIFF
--- a/Xcodes/Backend/XcodeCommands.swift
+++ b/Xcodes/Backend/XcodeCommands.swift
@@ -34,16 +34,26 @@ struct XcodeCommands: Commands {
 
 struct InstallButton: View {
     @EnvironmentObject var appState: AppState
+    @State private var isLoading = false
+
     let xcode: Xcode?
-    
+
     var body: some View {
         Button(action: install) {
-            Text("Install")
-                .help("Install")
-        }
+            if isLoading {
+                ProgressView()
+                    .colorInvert()
+                    .controlSize(.small)
+            } else {
+                Text("Install")
+                    .textCase(.uppercase)
+                    .help("InstallDescription")
+            }
+        }.buttonStyle(AppStoreButtonStyle(primary: false, highlighted: false))
     }
-    
+
     private func install() {
+        isLoading = true
         guard let xcode = xcode else { return }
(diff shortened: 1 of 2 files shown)


Xcodes/Backend/XcodeCommands.swift
```diff
@@ -34,16 +34,26 @@ struct XcodeCommands: Commands {
 
 struct InstallButton: View {
     @EnvironmentObject var appState: AppState
+    @State private var isLoading = false
+
     let xcode: Xcode?
-    
+
     var body: some View {
         Button(action: install) {
```
MattKiazyk, line 42: Can we reuse `ProgressButton` that we have already (2FA/SMS windows)

MattKiazyk (changes requested):
Thanks @chickdan !

Can we switch it to use `ProgressButton` that we already have?

Xcodes/Backend/XcodeCommands.swift
chickdan: Updated!

MattKiazyk (changes requested):

Xcodes/Backend/XcodeCommands.swift
```diff
@@ -34,16 +34,22 @@ struct XcodeCommands: Commands {
...
-        Button(action: install) {
+        ProgressButton(isInProgress: isLoading) {
+            install()
+        } label: {
             Text("Install")
-                .help("Install")
-        }
+                .textCase(.uppercase)
+                .help("InstallDescription")
+        }.buttonStyle(AppStoreButtonStyle(primary: false, highlighted: false))
```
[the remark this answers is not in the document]
chickdan: Good catch! Updated.

MattKiazyk (approved):

NSExceptional:
GitHub pro tip @chickdan: if you put something like `fixes #9999` in the PR title, it'll automatically close the issue when the PR is merged :)

PR #449, opened 2023-11-26
Support new Xcode 15 String Catalog

egesucu:
Xcode 15 supports String Catalog which improves the application localizations & adding new translations would be easier on every language.


egesucu:
Rebase has been implemented. For the case of strings, I've converted the main branch again and get the updated xcstrings file to update my branch and then rebase it.

PR #462, opened 2023-12-19
Add default phone number selection when there is only one

redryerye:
When I only have one trusted phone number and see the number selection view, I often get confused since it is hard to distinguish a tappable section in the list without a row highlight.
This change will select a phone number as a default when only one option exists.

|Single phone number |Multiple phone number |
|-|-|
|Screenshot 2023-12-20 at 0 44 13|Screenshot 2023-12-20 at 0 43 32|

|Single phone number with row selected |Multiple phone number with row selected |
|-|-|
|Screenshot 2023-12-20 at 0 43 06|Screenshot 2023-12-20 at 0 44 41|


MattKiazyk (approved):
Thanks so much @redryerye!

PR #470, opened 2024-01-05
Handle preference alerts from the PlatformsListView

chickdan:
The expected alert when choosing to delete a platform was not appearing. If I opened the `PreferencesView` and moved it off to the side I can see the alert appear over it when clicking the delete button. After adding this change I was able to see the alert in the `PlatformsListView`.

### Before:
Screenshot 2024-01-04 at 10 07 14 PM

### After:
Screenshot 2024-01-04 at 10 06 20 PM


MattKiazyk:
Thanks @chickdan - short story is originally I had it in Settings, but forgot to move the alert with it!

PR #482, opened 2024-01-20
Update zh_Hans Localizations. And more.

megabitsenmzq:
1. Add new Localizations. 

- "Alert.CancelInstall.Runtimes.Title"
- "Alert.DeletePlatform.PrimaryButton"
- "Alert.DeletePlatform.Title"
- "Perform post-install steps"
- "Error"
- "Open In Rosetta"
- "Platforms"
- "PlatformsList.Title"

2. Fix some old Localizations. All contains "前置步骤" -> "安装后准备步骤".

3. Remove old or useless Localizations.

4. ~~Create a file "String+.swift" to remove preview strings from localizations. Make it easier to work on translation.~~

5. Use Text(verbatim: ) to remove preview strings from localizations. Make it easier to work on translation.


Xcodes/Frontend/Common/String+.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+//  Created by Jinyu Meng on 2024/01/20.
+//  Copyright © 2024 Robots and Pencils. All rights reserved.
+//
+
+import Foundation
+
+extension String {
+    // Declare String as String explicitly. Prevent it from being recognized as a LocalizedStringKey.
+    var hideInLocalizations: String {
```
MattKiazyk, line 13: Curious on the benefit of adding this? Is there a places where we would want to do this vs just allowing it to go to the LocalizedStringKey?

MattKiazyk (commented):
I'm not seeing the value in adding in the `.hideInLocalizations` unless I'm missing something? Can you elaborate?

Xcodes/Frontend/Common/String+.swift
megabitsenmzq: If you check the other files I modified to add ".hideInLocalizations", you will notice that they are all in previews. These strings are only for preview purposes and don't need to be translated. But they will appear in xcstrings after compiling. So people who do the translation might be confused by this. Actually, I first translated all of them, then found out there was no need to do that.

MattKiazyk (changes requested):
In researching other localization issues - I came across `Text(verbatim:`
 https://developer.apple.com/documentation/swiftui/text/init(verbatim:) 

Which is a more simpler solution for what you're trying to accomplish so that it doesn't go into the .xcstrings file. 

Can you remove your `hideInLocalizations` variable and replace with verbatim where you want? Thanks

megabitsenmzq:
Got it! I will try it this weekend.

megabitsenmzq:
I found that although Text support initializes with init(verbatim: ), Picker or TextField don't have alternatives for this.

megabitsenmzq:
@MattKiazyk I removed "String+" and modified all Texts to use init(verbatim: ). For "Picker" and "TextField" which are not supporting init(verbatim: ), I gave them labels with Texts. Please check it out.

MattKiazyk:
@megabitsenmzq thanks so much for those changes - can you just rebase and I'll get that merged in

megabitsenmzq:
> @megabitsenmzq thanks so much for those changes - can you just rebase and I'll get that merged in

Done!

PR #521, opened 2024-03-08
Improve signing experience when using non-english keyboard

LiYanan2004:
## Context

Sometimes, we can not type password when we forgot to switch back to English keyboard.

## Fix

It may related to a focus issue, so I added a `.focused` modifier to resolve the issue.

I think that might because when `SecurityField` gets focus, it automatically switches to English while in process, `SecurityField` loses the focus ( not sure why ). 

By implementing the fix, it works correctly in any situation.

## Comparison 

Here is a comparison video for you to checkout what I have resolved.

https://github.com/XcodesOrg/XcodesApp/assets/37542129/ddca0d7b-aec3-40d0-a8c3-1b2cef3e8800

### DIFF
--- a/Xcodes/Frontend/SignIn/SignInCredentialsView.swift
+++ b/Xcodes/Frontend/SignIn/SignInCredentialsView.swift
@@ -1,9 +1,14 @@
 import SwiftUI
 
 struct SignInCredentialsView: View {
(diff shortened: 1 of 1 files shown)


MattKiazyk:
Thanks!

PR #522, opened 2024-03-08
Resolve concurrency check warning on Swift 5.10

LiYanan2004:
Swift 5.10 introduces stricter concurrency check which adds one more warning **`Reference to captured var 'progress' in concurrently-executing code; this is an error in Swift 6`**.

Because progress update is based on Notification and **operates on `OperationQueue.main`**, so we don't need to worry about data races.

So I opt-out concurrency check here by using `nonisolated(unsafe)`

### DIFF
--- a/Xcodes/Backend/Environment.swift
+++ b/Xcodes/Backend/Environment.swift
@@ -116,7 +116,8 @@ public struct Shell {
         return AsyncThrowingStream<Progress, Error> { continuation in
  
             Task {
-                var progress = Progress()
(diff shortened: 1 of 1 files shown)


MattKiazyk:
@LiYanan2004 can you clean up this PR so it just includes the concurrency check since the other change was already merged in via https://github.com/XcodesOrg/XcodesApp/pull/521

LiYanan2004:
@MattKiazyk Hi, thanks for mentioning this. It's my fault. I have cleaned the PR to only contains swift-5.10 concurrency fix.

PR #523, opened 2024-03-08
fix runtime not appearing when multiple runtimes exist for same build

MattKiazyk:
Fixes issue where iOS 17.4 runtime was not showing on Xcode 15.3

This was because we were grabbing only the first option (naïvely thinking that there would be a singular platform per release) and instead set the `sdkToSimulatorMappings` to an array.

This allows 17.4 to show properly.


rtharston:
Wow, nice timing on this!
I was going to report this but then I decided to fork and try and fix it myself instead of just logging an issue. Then I built locally and the problem was gone. Lo and behold, you merged this _right_ before I forked! 😁

PR #524, opened 2024-03-08
Release notes button isn't obvious

rtharston:
I wanted to see the release notes for Xcode 15.3 but I didn't see a button in the info pane, so this being an open source project I forked the repo to add a button.

Then when I looked at the code I realized the 🔗 button _is_ the release notes button I was planning on adding.

Now I know what the button does, but I obviously didn't understand it at first glance, and I didn't hover long enough to see the help tip either. So instead of adding a Release Notes button this PR replaces the 🔗 with text, to make it more discoverable

I don't know if this is the best solution, but I thought I'd share the idea and see what you think.
This approach also adds one more string to localize. I was able to copy out the pertinent bit of the ReleaseNotes.help string for a few of the romantic and germanic languages, but there are a few others left to do. I could look into doing those too if we decide to move forward with the text idea. (Or just use the help text directly, including the "see" part and not add a new string at all.)

PR
Screenshot 2024-03-08 at 1 24 37 PM

Released
Screenshot 2024-03-08 at 1 25 32 PM

### DIFF
--- a/Xcodes/Frontend/InfoPane/ReleaseDateView.swift
+++ b/Xcodes/Frontend/InfoPane/ReleaseDateView.swift
@@ -15,18 +15,16 @@ struct ReleaseDateView: View {
         if let date = date {
            
                 VStack(alignment: .leading) {
-                    HStack {
-                        Text("ReleaseDate")
-                            .font(.headline)
-                        Spacer()
-                        if let url {
-                            ReleaseNotesView(url: url)
-                        }
-                    }
+                    Text("ReleaseDate")
+                        .font(.headline)
                     
                     Text("\(date, style: .date)")
                         .font(.subheadline)
                   
+                    if let url {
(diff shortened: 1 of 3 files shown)


MattKiazyk:
🤔 agree the 🔗 icon may not be (at first glance) easy to understand. I was trying to keep it clean, so I think having that full release button goes against that. 

Perhaps changing the icon - to a 🗒️  or something would help?

rtharston:
I agree that the full text isn't as clean, so I understand why you went with the 🔗. I'm not sure if any other icon would be any more obvious, but the 🗒️ might be.

I also considered making the 'Release Notes' text smaller and removing the button border so it would have a cleaner look, but I'm not sure if that would be any better since it would then just look like text. Do you think underlined text would look clean enough? I'm not sure if I like the idea of using web style in a native app though. 🤔

PR #526, opened 2024-03-12
Trailing Icon for Filter Button

chickdan:
The label icon for the Xcode release filter jumped around when the filter was changed. This PR creates a new style that makes the icon trailing so that it stays in one place as the text changes.

### old:
old

### new:
No Xcode Selected


MattKiazyk:
Thanks again @chickdan !

PR #527, opened 2024-03-12
Remove outdated version checks

chickdan:
With macOS 13 being the minimum deployment target, some version checks are unnecessary. There are two other version checks skipped in this PR since they are removed in #526


chickdan:
Updated!

PR #530, opened 2024-03-16
replace default icon in info page

huihuisang:
Replace the default icon in info page.

**Before:**
image

**After**
image


MattKiazyk:
Thanks @huihuisang

PR #531, opened 2024-03-16
Replace AttributedText with markdown to fix the layout issues

huihuisang:
The height of AttributedText will change when switch difference tab in preference. 

**Screenshot**
Screenshot 2024-03-16 at 22 45 24

Screenshot 2024-03-16 at 22 44 13

**Solution**
Use Text and markdown format to replace of AttributedText.


MattKiazyk:
Thanks again @huihuisang

PR #535, opened 2024-03-17
fix: sidebar missing text on MacOS 14.4

MattKiazyk:
LOL. I love SwiftUI sometimes!

Fixes #534 

| Padding > 7 or nil | Other values |
|--------|--------|
| DUqCkUf5h1L2y9ZE4hxnJ87a | vTTmKEEyTRdlLrzfp8elDeoR |

PR #538, opened 2024-03-17
fix: xcode list row text hidden on macOS 14.4

### LINKED ISSUE #537: The layout is broken after upgrade to Version 2.1.1
**Describe the bug**
The layout of the list of the XCode versions are broken as shown below in the Screenshot section.

**To Reproduce**
Install XCodes from the Github release to my MacBook Pro, which installed macOS Sonoma 14.4 with Chinese Simplified lanaugage as System Language.

**Expected behavior**
A not broken layout, I suppose 🤣 
Bue seriously, the input search is so short, it must be broken, and there should be text of the XCode version number in each list item.
I think this is some special thing happend with SwiftUI and Chinese translation, but I can't be sure. I'll switch my System Language to English later just to test it out.

**Screenshots**
CleanShot 2024-03-17 at 19 42 51@2x

**Version**
 - OS: macOS Sonoma 14.4
 - Xcodes: 2.1.1

vvisionnn:
fixed #537
seems there is a pull request (#535) fixed this bug, but I think this PR is a more reasonable way to have this fixed instead of have a padding less than 8

@MattKiazyk and thanks for the quick responding!

another tip: while I try update the list using `ScrollView` + `LazyVStack` + `ForEach` should be ok compare to using `List` directly, but this will cause more unexpected error like the context menu only triggerred on the text area, so I think may be this  should be an optimization in the future

### Before
Screenshot 2024-03-17 at 20 51 04
### After
Screenshot 2024-03-17 at 20 50 56

### DIFF
--- a/Xcodes/Frontend/XcodeList/XcodeListView.swift
+++ b/Xcodes/Frontend/XcodeList/XcodeListView.swift
@@ -63,11 +63,11 @@ struct PlatformsPocket: View {
                 Image(systemName: "square.3.layers.3d")
                     .font(.title3.weight(.medium))
                 Text("PlatformsDescription")
+								Spacer()
             }
             .font(.body.weight(.medium))
             .padding(.horizontal)
             .padding(.vertical, 12)
-            .frame(maxWidth: .infinity, alignment: .leading)
             .background(.quaternary.opacity(0.75))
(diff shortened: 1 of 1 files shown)


MattKiazyk:
Thanks @vvisionnn

PR #539, opened 2024-03-18
Resolve content clipping inside Xcode detail view

LiYanan2004:
## Problems 

Inside Xcode's detail view, I noticed there is a hard clip, which is annoying to me 🤣.

## Fix 

Replace `padding` with `contentPadding` (custom modifier using safeArea) to avoid ScrollView's clipping.

## Comparison

https://github.com/XcodesOrg/XcodesApp/assets/37542129/8151dc34-858f-428f-bd27-1f4abfd57f01

Much better.

### DIFF
--- a/Xcodes/Frontend/MainWindow.swift
+++ b/Xcodes/Frontend/MainWindow.swift
@@ -41,7 +41,7 @@ struct MainWindow: View {
                     UnselectedView()
                 }
             }
-            .padding()
+            .contentPadding()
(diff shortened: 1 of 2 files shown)


LiYanan2004:
Not sure why the tests are failed. Seems to be okay.

LiYanan2004:
@MattKiazyk Looks good. But older OS deserve this margin as well. So implemented it again. Please check it out😁

PR #540, opened 2024-03-20
InfoPane UI Tweaks

marlonjames71:
### Main Goal

I wanted to update the way the platform views looked when it comes to the position of the "Install" button. It sits below the platform title creating extra space unnecessarily in my opinion. This looked confusing to me because I thought there might have been missing text. My change in regards to that puts the "Install" button inline with the platform title. Benefits include:

- uniform look for all detail views
- doesn't look like something should be there and is missing—text, for example
- that view doesn't change size when "Install" buttons aren't available
- also just looks cleaner

### Other Changes

- **ScrollView Content Clipping:**
The InfoPane's scrollable content was getting clipped due to the padding that was added to the InfoPane. I fixed this with a new SwiftUI modifier: `contentMargins`. I use an availability check first, so old clipped code will remain for anyone using macOS older that v14. This is nice since the added code is very minimal. The only change from the old is that the padding was moved into the InfoPane's file making it responsible for its own layout.

- **Indentation Inconsistencies:**
I usually reindent files to fix any indentation that needs fixin'. 

- **README Update:**
Because there were changes made to the UI in the info pane, I also went ahead and updated the images that show in the README file. Hope that's okay.

### Here are some images
| **Old** | **New** |
| - | - |
| CleanShot 2024-03-19 at 22 39 41@2x | CleanShot 2024-03-19 at 22 39 19@2x


MattKiazyk:
Thanks @marlonjames71 - tweaks looks awesome!

PR #544, opened 2024-03-24
Adds margins to InfoPannel for older OS

LiYanan2004:
Previous PR that has been closed without merging: #539 

Based on #540, older OS still using padding around ScrollView which results in a hard clip. Technically, it's possible to add margins inside ScrollView, so, of course, older OS deserves that margin🤣. 

The code here uses safe area to add margins which might be the same with the new margin API, but supports older OS.


MattKiazyk:
@LiYanan2004 can you show me what this looks like on MacOS < 14? Thanks

LiYanan2004:
Unfortunately, I have no device running macOS 13. But that piece of code should look the same as the latest OS.

PR #547, opened 2024-04-05
Set explicit SwiftPolyglot tag and update clone path

roddymunro:
I noticed there wasn't an explicit tag for `SwiftPolyglot`. I wanted to make sure your PRs are not affected by any potential breaking changes in the future, so setting the tag will prevent any issues before you get the chance to upgrade.

### DIFF
--- a/.github/workflows/xcstrings.yml
+++ b/.github/workflows/xcstrings.yml
@@ -14,9 +14,9 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Clone SwiftPolyglot
-      run: git clone https://github.com/appdecostudio/SwiftPolyglot.git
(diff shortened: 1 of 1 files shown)


.github/workflows/xcstrings.yml
```diff
@@ -14,9 +14,9 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Clone SwiftPolyglot
-      run: git clone https://github.com/appdecostudio/SwiftPolyglot.git
+      run: git clone https://github.com/appdecostudio/SwiftPolyglot.git --branch 0.3.1 ../SwiftPolyglot
```
nevack, line 17:
```suggestion
      run: git clone --branch 0.3.1 -- https://github.com/appdecostudio/SwiftPolyglot.git ../SwiftPolyglot
```

roddymunro: @nevack Thanks, done

PR #560, opened 2024-05-14
chore: lossless png compression

mikeirvingweb:
Lossless PNG image compression

smaller downloads, smaller repo, smaller app size


chickdan:
It's subtle but the colors of both the Robots & Pencils logo as well as the app icon are muted after being compressed.

mikeirvingweb:
I think that's due to an opacity thing.. obvious on the app icon.

I will make a further commit at some point, rolling back those two

mikeirvingweb:
@chickdan I have rolled back the app icons, including icon.png in the root, and the beta app icon, and the R&P graphic.

so there's much less in the PR now, but hopefully what is left is fine.

MattKiazyk:
👍

PR #590, opened 2024-07-04
[Issue-578] - adjust AuthOptionsResponse parser decode

anettodev:
When the user has a 2FA with security keys (yubikey), the AuthOptionsResponse cannot be parsed because `securityCode` attribute is not optional and the `AuthenticationError.accountUsesUnknownAuthenticationKind` is never shown.

The correct authType for security keys is not implemented and needs to check more the documentation for this.

PS: also, I have updated some missing pt-br localization strings @MattKiazyk

Before:
Screenshot 2024-07-04 at 00 02 09

After
Screenshot 2024-07-04 at 00 02 09

### DIFF
--- a/Xcodes/AppleAPI/Sources/AppleAPI/Client.swift
+++ b/Xcodes/AppleAPI/Sources/AppleAPI/Client.swift
@@ -331,15 +331,15 @@ public enum TwoFactorOption: Equatable {
 public struct AuthOptionsResponse: Equatable, Decodable {
     public let trustedPhoneNumbers: [TrustedPhoneNumber]?
     public let trustedDevices: [TrustedDevice]?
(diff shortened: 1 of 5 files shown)


anettodev:
If someone can provide a better doc about `idmsa.apple.com/appleauth/auth`, `"X-Apple-ID-Session-Id"`, `"scnt"`, `"X-Apple-Widget-Key"`, I can dig a little more.

Already checked the 
https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api

MattKiazyk:
@anettodev support for authentication keys have been added in  https://github.com/XcodesOrg/XcodesApp/pull/617 so I'm going to close this PR.

Issue #572, opened 2024-06-10
The filter button disappears after tapping it.


oliverfoggin:
**Describe the bug**
After tapping the filter button it disappears.

**To Reproduce**
Open the app.
Tap the filter button.

Hiding and showing the side bar doesn't seem to fix this.
Quitting and opening doesn't have an effect either.

**Expected behavior**
The button remains so that I can change the filter again.

**Screenshots**
<img width="912" alt="image" src="https://github.com/XcodesOrg/XcodesApp/assets/1813244/28c70361-feab-4f43-8cb7-51d2dc3451c7">

**Version**
 - OS: Sonoma 14.5
 - Xcodes: 2.1.2 (26)

oliverfoggin:
NVM...

Changing the size of the sidebar fixes it. Would be nice if this didn't happen though.

Issue #550, opened 2024-04-10
Unable to reset filter; stuck on only seeing installed platforms.


HolyBiscuits:
**Describe the bug**

The UI doesn't allow me to disable the filter that only shows me installed XCode platforms. It seems that the filter icons have disappeared, and I can't get it to revert to its original state through settings or moving the application to trash and reopening the original downloaded file from the zip folder.

At this point, I've installed v1.10, so I'm still able to use the functionality of XCodes that I was looking for. I appreciate any help to restore the UI to the default state on the latest version!

**To Reproduce**

I'm not exactly sure; I got started with this program, messed around with the filters a bit, and ended up stuck on only being able to access my currently installed XCode versions.

**Expected behavior**

I expect for icons to have appeared allowing me to disable the filter. If there is a way to revert back to the original view, it is not clear to me how I am supposed to do so.

**Screenshots**

<img width="895" alt="Screenshot 2024-04-10 at 12 59 43 PM" src="https://github.com/XcodesOrg/XcodesApp/assets/66638350/c73d2514-debe-493f-a049-6cfa0ed1e9f3">

**Version**
 - OS: Sonoma 14.4.1
 - Xcodes: v2.1.2

MattKiazyk:
I believe MacOS will shove those in the `>>`  sometimes when it thinks it can't draw it on the nav bar

![lnuwxOS9Y93WIBx3kzDGhNxQ](https://github.com/XcodesOrg/XcodesApp/assets/1119565/01efabb0-666b-4299-9ef6-98a1101abe19)

iosdevben:
You are correct @MattKiazyk.

They now appear in that "overflow" menu for me.

![image](https://github.com/XcodesOrg/XcodesApp/assets/824608/942ad5cc-4041-4c23-ac01-8ed88626c4ce)

laveur:
I made the same mistake, and the options in the overflow menu are greyed out and can not be unselected.

laveur:
As a hack if you are stuck like I was: `defaults write com.xcodesorg.xcodesapp xcodeListCategory ""` will fix you.

Issue #498, opened 2024-02-04
Stuck on beta filter


AzSiAz:
**Describe the bug**
I can't filter Xcode release and I am stuck on beta filter. Based on latest release screenshot it also seem filters are not at the correct place.

**To Reproduce**
I have no idea, only thing I remember is clicking on filter sorry 😅 

**Expected behavior**
Can filter

**Screenshots**
<img width="1512" alt="Capture d’écran 2024-02-04 à 17 01 35" src="https://github.com/XcodesOrg/XcodesApp/assets/7109680/27f6c6e9-b0cd-4d42-903b-5f2952420ee9">

**Version**
 - OS: 14.3
 - Xcodes: 2.0.2 (23)

AzSiAz:
Ah, I should have tried to resize the left sidebar. This bug seem to only be when there is not enough space to show activated text filter

PR #591, opened 2024-07-05
Remove labels from filter button on toolbar

CatMe0w:
This is related to #498, #550, #572.

With the default width of the sidebar, clicking the filter button on the toolbar will make the button disappear from the sidebar area, ending up in the rightmost corner:
image
image

Technically, it works as intended, but it's very confusing from the perspective of users: it looks like the buttons just disappear. Dragging the sidebar to make it wider can reveal the buttons, but it's not intuitive. Actually, I got stuck here for a few days, trying to figure out where my beta versions went.

I think we might avoid the overflow entirely, and I tried some different approaches like using abbreviations instead of full text such as "Release only," but the overflow still happens and does not look good:
image

Finally, I found that completely removing the labels could be the most ideal solution. A filled, highlighted button with floating help text could be intuitive enough and won't break the consistency of the UI:
image
image

I hope this change improves the user experience and keeps the UI clean and functional.

### DIFF
--- a/Xcodes/Frontend/XcodeList/MainToolbar.swift
+++ b/Xcodes/Frontend/XcodeList/MainToolbar.swift
@@ -33,17 +33,17 @@ struct MainToolbarModifier: ViewModifier {
                 switch category {
                 case .all:
                     Label("All", systemImage: "line.horizontal.3.decrease.circle")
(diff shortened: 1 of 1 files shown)


CatMe0w:
This pull request is now obsolete with the v3.0 release.
image

PR #603, opened 2024-08-17
Update label for Cancel Install Button

LiYanan2004:
## Before 
截屏2024-08-17 11 20 01

## After 
截屏2024-08-17 11 34 26

This gives more clarity to the button intent and better matches other types of context menus.


MattKiazyk (approved):
Thanks! Looks awesome

PR #621, opened 2024-10-11
Terminate Xcodes app after last window closed

### LINKED ISSUE #620: Xcodes does not quit after last window closed
**Describe the bug**

We do not exit Xcodes after the main window is close.

So when user click Xcodes.app again. No window will be created.

Instead they need to force quit and open it again or first active Xcodes.app and then click File -> New Xcodes Window / ⌘ N

The fix is easy by adding the following code to `AppDelegate` class

```swift
func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool {
    true
}
```

**To Reproduce**

- Close the last window of Xcodes.app
- Click Xcodes.app and no new window is created.

**Expected behavior**

Quit Xcodes so that a new click will reopen a Xcodes window.

**Version**
 - OS: macOS 15
 - Xcodes:

Kyle-Ye:
Close #620


MattKiazyk (changes requested):
@Kyle-Ye Thanks for the PR!

I'm ok adding this in if a setting gets put inside of `Setting > General` with the default being not to do this. Thanks

Kyle-Ye:
Got it. I can accept adding it into Setting > General with the default being not to do this. So it at least fix my side.

But also found the root cause is that after we migrate into SwiftUI lifecycle. We get automatic multi window support for the main Window using `WindowGroup`.

Changing into macOS 13's `nonisolated public init(_ titleKey: LocalizedStringKey, id: String, @ViewBuilder content: () -> Content)` API will disable the multi window support for the main Window and will also fix the issue I mentioned above.

```diff
-        WindowGroup("Xcodes") {
+        Window("Xcodes", id: "main") {
```

Does supporting multi main window a feature we'd like to support? If not, I'd prefer we just switch to the new API and kill the multi window feature.

image

MattKiazyk:
@Kyle-Ye I'm for not having multi window as we don't need it, however I don't want Xcodes to close on X with this change to `Window` unless the user wants it to.

Kyle-Ye:
> I'm for not having multi window as we don't need it

Got it. But if we do not want multi window support. The `WindowGroup` -> `Window` change is needed.

> however I don't want Xcodes to close on X
 
If the app does not have multi windows support, the default behavior is terminate the app after last window is closed. But we can override the behavior with override `applicationShouldTerminateAfterLastWindowClosed` and return false here.

eg.
1️⃣. Multi Windows support: (The current `main` branch behavior)
- 1 window -⌘-N-> 2 window -⌘-N-> 3 window
- 1 window -⌘-W-> 0 window(Still alive) -⌘-N-> 1 window
- 1 window -⌘-Q-> terminate app

2️⃣. Multi Windows support with the option to terminate on when last window is closed: (`applicationShouldTerminateAfterLastWindowClosed` return the value we set on Settings)
- 1 window -⌘-N-> 2 window -⌘-N-> 3 window
- 1 window -⌘-W-> 0 window(setting option value is true) and terminate / 0 window and not terminate -⌘-N-> 1 window
- 1 window -⌘-Q-> terminate app

3️⃣. Without multi Windows support: (`WindowsGroup` -> `Windows`)
- 1 window (No ⌘ N support)
- 1 window -⌘ W-> 0 window and terminate app by default
- 1 window -⌘-Q-> terminate app

4️⃣. Without multi Windows support and do not terminate: (`WindowsGroup` -> `Windows` && `applicationShouldTerminateAfterLastWindowClosed` return false)

- 1 window (No ⌘ N support)
- 1 window -⌘ W-> 0 visible window and does terminate the app (Still have dot on dock since the window is not actually closed in this situation) -> click the app icon to restore the closed window
- 1 window -⌘ Q-> terminate app

The round 1 review suggestion seems is using method 2️⃣.

And the latest commits of the PR is using method 3️⃣.

IIUC, you are suggesting we should use method 4️⃣?

MattKiazyk:
> 4️⃣. Without multi Windows support and do not terminate: (WindowsGroup -> Windows && applicationShouldTerminateAfterLastWindowClosed return false)

1 window (No ⌘ N support)
1 window -⌘ W-> 0 visible window and does terminate the app (Still have dot on dock since the window is not actually closed in this situation) -> click the app icon to restore the closed window
1 window -⌘ Q-> terminate app

Correct. 4️⃣ would be the preference with `applicationShouldTerminateAfterLastWindowClosed` be a user setting that if users wanted it to terminate on close they could do so.

Kyle-Ye:
> > 4️⃣. Without multi Windows support and do not terminate: (WindowsGroup -> Windows && applicationShouldTerminateAfterLastWindowClosed return false)
> 
> 1 window (No ⌘ N support) 1 window -⌘ W-> 0 visible window and does terminate the app (Still have dot on dock since the window is not actually closed in this situation) -> click the app icon to restore the closed window 1 window -⌘ Q-> terminate app
> 
> Correct. 4️⃣ would be the preference with `applicationShouldTerminateAfterLastWindowClosed` be a user setting that if users wanted it to terminate on close they could do so.

Got it. But the default OS behavior would be close (The same as we return `true` for `applicationShouldTerminateAfterLastWindowClosed`) - see method 3️⃣.

So I should use 4️⃣ with `applicationShouldTerminateAfterLastWindowClosed` be a user setting and **the default value set to false**, right?

Kyle-Ye:
Update method 4️⃣ implementation with the `applicationShouldTerminateAfterLastWindowClosed` settings.

I am not familiar with all language Xcodes support. (Only add localization for language I know.)

Do I need to use some AI service/translator to translate the new LocalizedString keys here? I can see `XCStrings Validation / test (pull_request)` CI is failing here.

MattKiazyk:
@Kyle-Ye for the keys you've added, please paste the English translation to each language in the value. 

The app isn't hooked up to any automated translation service, so we rely on the community to see the English in their app and make PR's. 

Without adding the default translation other languages see the key as the value.

Kyle-Ye:
> @Kyle-Ye for the keys you've added, please paste the English translation to each language in the value.
> 
> The app isn't hooked up to any automated translation service, so we rely on the community to see the English in their app and make PR's.
> 
> Without adding the default translation other languages see the key as the value.

Done. Also fix another tiny issue for pl Localization.

Xcodes/Frontend/Preferences/AdvancedPreferencePane.swift
```diff
@@ -151,6 +151,11 @@ struct AdvancedPreferencePane: View {
                 }
             }
             .groupBoxStyle(PreferencesGroupBoxStyle())
+            
+            GroupBox(label: Text("Misc")) {
```
MattKiazyk, line 155: @Kyle-Ye thoughts on putting it under general vs advanced tab? I almost feel this is a pretty regular MacOS feature and I would lean towards putting it under general

Kyle-Ye: Done

MattKiazyk (approved):
@Kyle-Ye thanks for your time and going back and forth on this!

PR #638, opened 2024-10-28
SRP Login works now

abiligiri:
- Switch to use https://github.com/adam-fowler/swift-srp with some modifications that are local
  - Pad g value to equal size of N while calculating clientProof
- Use SHA256(plain-text-password) while computing key using PBKDF2
- Added a unit test with some sample values

Fixes https://github.com/XcodesOrg/XcodesApp/issues/630


fiznool:
Personally, I like the approach of using a well-tested library for SRP, and the upstream library seems to provide this. So I would be in favour of this implementation vs #639 

If I can make a suggestion, forking adam-fowler/swift-srp and publishing as a separate package with the necessary changes (xcodes-srp ?) would be my preference for the following reasons:

- any changes to the upstream code can be more easily incorporated into a fork
- the shared package can be consumed by the Xcodes app and cli

abiligiri:
Note that both options use a library for SRP. This uses https://github.com/adam-fowler/swift-srp/blob/main/Sources/SRP/client.swift and #639 fixes the srp code that was already in repo most likely from https://github.com/Bouke/SRP. I see that Adam Fowler implementation is still active. Maybe that is another reason to use it.

@MattKiazyk  I was able to download and install Xcode 16.1.0. Not sure what you mean by download doesn't work.

abiligiri:
May I also request merging this for the release https://github.com/XcodesOrg/XcodesApp/pull/632

abiligiri:
I cleaned this up a bit. Using patched swift-srp from https://github.com/abiligiri/swift-srp and removed local copy. I have also opened MR for upstream https://github.com/adam-fowler/swift-srp/pull/13

MattKiazyk:
@abiligiri hope you don't mind I clean up some extra prints that were in there as well as moved the `swift-srp` dependency to an xcodes fork. I copied your changes from your branch into main on our fork.

PR #639, opened 2024-10-28
Fix SRP Login

abiligiri:
- Derive key using SHA25(plain-text-password) via PBKDF2
- Pad g to size N while calculating M
- Handle password sent as base64 encoded data and account for absence of username

Fixes https://github.com/XcodesOrg/XcodesApp/issues/630


abiligiri:
Closing this in favor of #638

PR #640, opened 2024-10-29
Support SRP Login

MattKiazyk:
Big thanks to @abiligiri for figuring out the hard stuff


Xcodes/AppleAPI/Package.swift
```diff
@@ -1,24 +1,26 @@
...
+    platforms: [.macOS(.v11)],
     products: [
         // Products define the executables and libraries a package produces, and make them visible to other packages.
         .library(
             name: "AppleAPI",
             targets: ["AppleAPI"]),
     ],
-    dependencies: [],
+    dependencies: [
+        .package(url: "https://github.com/xcodesOrg/swift-srp", branch: "main")
```
tsdevshop, line 16: It might be better to depend on a particular commit rather than `main`, as this could cause issues if `swift-srp` introduces a change which is not compatible with the current app codebase.

Or does it make sense to package this up as a separate SRP package, something like `icloud-srp`? I imagine this will be useful for other people who need to authenticate with Apple's iCloud servers in Swift.

PR #650, opened 2024-11-05
Fix handling of s2k_fo

### LINKED ISSUE #644: How to Fix - invalid username password when logging in
When logging in with the latest versions, you may receive an invalid username or password error even though it is correct. 

With the new [SRP ceremonial](https://blog.uniauth.com/what-is-secure-remote-password) it requires the client to send a salt and a precomputed password verifier value to the server. The server must store these values to later authenticate the client.

If your password is very old (pre SRP), it is likely that Apple servers did not receive these values for your password, hence can not authenticate you. 

Changing your password via https://account.apple.com/account/manage fixes the issue by updating these values to Apple's servers.

kabiroberai:
Apple IDs with "old" passwords (ie those set before a certain date, the cutoff for which I'm not 100% certain) use the `s2k_fo` protocol rather than `s2k`. This legacy protocol involved an extra step in the SRP handshake, where the password was hex-encoded before performing PBKDF2 to derive the shared secret. Luckily it's pretty trivial for us to handle this and fix the erroneous "incorrect username or password" errors that it resulted in.

Closes #644 — this is in fact the root cause behind that issue.

### DIFF
--- a/Xcodes/AppleAPI/Sources/AppleAPI/Client.swift
+++ b/Xcodes/AppleAPI/Sources/AppleAPI/Client.swift
@@ -54,7 +54,7 @@ public class Client {
                 let iterations = srpInit.iteration
                 
                 do {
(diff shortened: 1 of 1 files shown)


MattKiazyk:
Amazing!

PR #655, opened 2024-11-10
Support FIDO2 authentication with devices that don’t have a PIN code

MultiColourPixel:
This PR adds support for FIDO2 authentication with devices that don’t have a PIN code set.

This depends on a newer release of the `LibFido2Swift` library. 
At the time of writing `LibFido2Swift` only supports pin-less assertation on the main branch. As such this PR will have to wait until @kinoroy creates a new release on his side.

The button for submitting without a pin was added to be inline with the other buttons on the PIN screen.

Updated PIN Request flow

### DIFF
--- a/Xcodes/Backend/AppState.swift
+++ b/Xcodes/Backend/AppState.swift
@@ -357,7 +357,7 @@ class AppState: ObservableObject {
     
     var fido2: FIDO2?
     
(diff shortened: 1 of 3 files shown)


kinoroy:
Hey, thanks for this! I went and created a release [v0.1.3](https://github.com/kinoroy/LibFido2Swift/releases/tag/0.1.3) for LibFido2Swift with support for PIN-less assertions

PR #686, opened 2025-01-14
[Feature]Allow opening incompatible Xcode versions

ipavlidakis:
This revision, implements the mechanism that allows opening Xcode versions that are incompatible with the current macOS version. That applies on all Xcode versions that have been released with older or later macOS versions than the running one.

With this revision, by tapping `Open`, we check if the running major version is equal to the selected Xcode's required major version:
- if they are equal, we perform the `NSWorkspace` openApplication like usual
- if they are not, we construct a URL to the Xcode's binary (<Xcode Application Path>/Contents/MacOS/Xcode). Then we instantiate a new process and pass the binary path for execution (we use a Process rather than asking the NSWorkspace to open the binary, to avoid showing the terminal window).

### DIFF
--- a/Xcodes/Backend/AppState.swift
+++ b/Xcodes/Backend/AppState.swift
@@ -738,18 +738,64 @@ class AppState: ObservableObject {
     func open(xcode: Xcode, openInRosetta: Bool? = false) {
         switch xcode.installState {
         case let .installed(path):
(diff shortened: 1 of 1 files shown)


luey-p:
@MattKiazyk Instead of navigating to Xcode's binary to open the app, opening it directly through the Xcodes app for these unsupported Xcode versions would be amazing.

PR #746, opened 2025-09-02
Improve filtering UI/UX on Xcode List

LiYanan2004:
Filter controls should be compact in toolbar since the smaller room, and could be larger inside the app

Check out the new UX here

### Before
https://github.com/user-attachments/assets/6f6514da-0c2f-4acc-85b1-7b0f32585a01

### After
https://github.com/user-attachments/assets/736d41ce-1292-400b-abb3-1fe76666d406


MattKiazyk:
Thanks for this @LiYanan2004 - looks a lot cleaner

PR #770, opened 2025-10-13
TR Localization Update

egesucu:
This PR adds missing TR localizations


egesucu:
Semms like String validation checks missing localizations for other languages which is out of my scope :)

PR #778, opened 2025-11-12
fix: improve paste support for PIN code input fields

zhangqifan:
## Description
Improves the paste functionality for PIN code input fields used in SMS and 2FA authentication flows.

## Problem
When pasting a 6-digit verification code like "123456" into the PIN code input fields, only the last digit ("6") would appear in the first field. The other digits were lost.

## Solution
- Add detection for multi-character paste operations
- Filter pasted content to alphanumeric characters only
- Always start filling from the first field when pasting
- Optimize performance by updating all fields at once
- Support both numeric and alphanumeric codes (e.g., 123456, A1B2C3)

## Impact
This improvement benefits both:
- **SignInSMSView** - SMS verification code input
- **SignIn2FAView** - 2FA device verification code input

Both views share the same `PinCodeTextField` component, so the fix applies automatically to both flows.

## Testing
Tested with:
- ✅ Pasting 6-digit numeric codes (123456)
- ✅ Pasting with cursor in different positions (always fills from first field)
- ✅ Pasting codes with special characters (auto-filtered to alphanumeric only)
- ✅ Manual single-character input (unchanged behavior)

## Checklist
- [x] Code follows the project's style guidelines
- [x] Self-review of code completed
- [x] No new warnings or errors introduced
- [x] Changes tested with real authentication flow

### DIFF
--- a/Xcodes/Frontend/SignIn/PinCodeTextView.swift
+++ b/Xcodes/Frontend/SignIn/PinCodeTextView.swift
@@ -149,6 +149,34 @@ class PinCodeTextView: NSControl, NSTextFieldDelegate {
         
         let newFieldText = field.stringValue
         
+        // Handle pasting multiple characters (e.g., pasting "123456" from clipboard)
+        if newFieldText.count > 1 {
+            // Filter to alphanumeric characters only
+            let validCharacters = newFieldText.filter { $0.isLetter || $0.isNumber }
+            
+            // Always start from the first field and clear previous content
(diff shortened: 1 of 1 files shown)


SagarSDagdu:
@MattKiazyk This is solving a real pain-point in the app, it would be great if could merge this soon.

nixta:
Yeah. This is desperately needed please! When you click on the macOS-provided popup that's parsed your 2FA code from the incoming SMS message, macOS deletes the message (I think this is default behaviour). So you get the first digit, but then the other 5 digits are lost to you.

Issue #782, opened 2025-12-11
Clicking Install Apple Silicon Installs Universal Instead


spaus:
**Describe the bug**
When I filter on Apple Silicon builds, and click Install or Install Apple Silicon nothing happens. If I then filter on Universal, I see that the universal builds are installing.

**To Reproduce**
1. Filter on Apple Silicon
2. Select an Apple Silicon build
3. Click Install or Install Apple Silicon
4. Notice that nothing happens
5. Filter on Universal
6. Notice that the Universal builds are installing

**Expected behavior**
The Apple Silicon build installs

**Screenshots**

**Version**
 - OS: macOS 15.6
 - Xcodes: 26.1.1, 26.1, 26.0.1, 26.0

Issue #777, opened 2025-11-10
When I select "install apple silicon", nothing happens there, it instead installs universal


gsbernstein:
**Describe the bug**
Selecting "install apple silicon" instead installs the "universal" version (or at least appears to)

**To Reproduce**
1. select a row for an apple silicon version
2. select "install apple silicon"

**Expected behavior**
State for that apple silicon entry should change to installing, button should change to "cancel"

**Actual behavior**
Nothing happens on the apple silicon entry, instead the state changes to installing on the universal version above. When installation is complete, it only shows that the universal version is installed. After refreshing, it shows that both versions are installed.

Reproducibility: 100%

**Screenshots**

<img width="842" height="278" alt="Image" src="https://github.com/user-attachments/assets/dcfee45a-f664-4b79-8e38-cea299a0a881" />

https://github.com/user-attachments/assets/5e711e2a-cc94-4bdc-ac1f-98c6b3b3170a

When it's done:
<img width="384" height="110" alt="Image" src="https://github.com/user-attachments/assets/833f424b-646f-41c0-899a-32a48851fd30" />

After refreshing:
<img width="375" height="99" alt="Image" src="https://github.com/user-attachments/assets/7b4bfe8b-df0a-4155-92a0-848f8ff8b936" />

**Version**
 - OS: 15.6 (24G84)
 - Xcodes: 3.0.2 (24) (you should make this selectable in the About panel!)

esperpanorama:
I can reproduce this issue.

kapitoshka438:
It actually installs the correct variant of Xcode. It's just a visual bug, quite annoying.
```
lipo -info /Applications/Xcode-26.1.0.app/Contents/MacOS/Xcode 
Non-fat file: /Applications/Xcode-26.1.0.app/Contents/MacOS/Xcode is architecture: arm64
```

markst:
Same here, I clicked "Install Apple Silicon" but went ahead and started installing universal:

<img width="720" alt="Image" src="https://github.com/user-attachments/assets/98aeb0f4-a4ec-496f-9ba5-d34597bc3166" />

PR #787, opened 2026-02-02
Fix: Install Apple Silicon shows Universal as Installing

### LINKED ISSUE #776: ‘Install Apple Silicon’ installs Universal build instead of Apple Silicon-only version on macOS 26.1
Description

When I click “Install Apple Silicon” in the Xcodes GUI, it installs the Universal version of Xcode instead of the Apple Silicon-only build.
This happens consistently on macOS 26.1 with Xcodes 3.0.2 (34).

⸻

Steps to Reproduce
	1.	Launch Xcodes 3.0.2 on macOS 26.1 (Apple Silicon)
	2.	Select Xcode 26.1 (17B55)
	3.	Click “Install Apple Silicon”
	4.	Wait for authentication and installation to complete

Expected:
The Apple Silicon-only build of Xcode 26.1 should be installed (arm64-only).

Actual:
The Universal version (arm64 + x86_64) is installed instead.

⸻

Environment
	•	macOS: 26.1 (Apple Silicon)
	•	Xcodes version: 3.0.2 (Build 34)
	•	Xcode target: 26.1 (17B55)
	•	Reproducibility: Always
	•	Network: Normal (no proxy/VPN)

⸻

Additional Information
	•	The issue persists even after clearing cache and reinstalling Xcodes.
	•	The UI label correctly shows “Apple Silicon” in the platform selector, but the resulting installation is Universal.
	•	The problem does not occur when using the CLI tool (xcodes install 26.1 --platform=applesilicon).

Image
Image

william-laverty:
## Summary
When clicking 'Install Apple Silicon', the UI incorrectly shows the Universal version as 'Installing' instead of the Apple Silicon version.

## Root Cause
`setInstallationStep` used `version.isEquivalent()` which doesn't consider architectures. When both Universal and Apple Silicon versions exist with the same Xcode version, the first one found (Universal) was marked as installing.

```swift
// Before: Uses version equivalence (ignores architectures)
func setInstallationStep(of version: Version, to step: XcodeInstallationStep) {
    guard let index = self.allXcodes.firstIndex(where: { $0.version.isEquivalent(to: version) }) else { return }
    // ...
}
```

## Fix
Changed `setInstallationStep` to accept `XcodeID` instead of `Version`, enabling exact matching including architectures:

```swift
// After: Uses XcodeID equality (includes architectures)
func setInstallationStep(of xcodeID: XcodeID, to step: XcodeInstallationStep) {
    guard let index = self.allXcodes.firstIndex(where: { $0.id == xcodeID }) else { return }
    // ...
}
```

## Testing
1. Filter list to show only Apple Silicon builds
2. Select an Xcode version with both Universal and Apple Silicon variants
3. Click 'Install Apple Silicon'
4. Verify the Apple Silicon row shows as 'Installing', not the Universal row

Fixes #776, #777, #782

### DIFF
--- a/Xcodes/Backend/AppState+Install.swift
+++ b/Xcodes/Backend/AppState+Install.swift
@@ -104,7 +104,7 @@ extension AppState {
     private func downloadXcode(availableXcode: AvailableXcode, downloader: Downloader) -> AnyPublisher<(AvailableXcode, URL), Error> {
             self.downloadOrUseExistingArchive(for: availableXcode, downloader: downloader, progressChanged: { [unowned self] progress in
                 DispatchQueue.main.async {
-                    self.setInstallationStep(of: availableXcode.version, to: .downloading(progress: progress))
+                    self.setInstallationStep(of: availableXcode.xcodeID, to: .downloading(progress: progress))
                     self.overallProgress.addChild(progress, withPendingUnitCount: AppState.totalProgressUnits - AppState.unxipProgressWeight)
                 }
             })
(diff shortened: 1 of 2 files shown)


ddaddy:
Can we not get this merged? Xcodes is downloading Apple Silicon on my Intel Mac.

Issue #815, opened 2026-06-08
Crash upon launch on Mac OS 27 developer beta 1


xavier114fch:
**Describe the bug**
Xcodes 4.0.2 crashes when it is launched on Mac OS 27 developer beta 1.

**To Reproduce**
1. Launch Xcodes 4.0.2 from any places (Application folder, Spotlight etc.)
2. Crash immediately with Problem Report screen from Mac OS

**Expected behavior**
1. App launches successfully

**Screenshots**
<img width="1112" height="912" alt="Image" src="https://github.com/user-attachments/assets/b43bd54f-d4a0-4739-8def-c96b06eb4c3c" />

**Stack trace**
[trace.txt](https://github.com/user-attachments/files/28725224/trace.txt)

**Version**
 - OS: 27 DB1
 - Xcodes: 4.0.2

MattKiazyk:
Not even a few hours in!

@xavier114fch  I probably won't even look at it till at least next beta unfortunately. I've been burnt before trying to fix too early where they fix by the next beta.

bennettp123:
until then, [xcodes-cli](https://github.com/XcodesOrg/xcodes) is working just fine

fbarbat:
@bennettp123 that's only if you have it already installed, right? Because I can't install it using brew.
```
Error: You are using macOS 27.
We do not provide support for this pre-release version.
```

bennettp123:
> [@bennettp123](https://github.com/bennettp123) that's only if you have it already installed, right? Because I can't install it using brew.
> 
> ```
> Error: You are using macOS 27.
> We do not provide support for this pre-release version.
> ```

After reinstalling homebrew, I was able to get it using `brew install xcodes`.

MattKiazyk:
> > [@bennettp123](https://github.com/bennettp123) that's only if you have it already installed, right? Because I can't install it using brew.
> > ```
> > Error: You are using macOS 27.
> > We do not provide support for this pre-release version.
> > ```
> 
> After reinstalling homebrew, I was able to get it using `brew install xcodes`.

You should use `brew install xcodesorg/made/xcodes`

fbarbat:
@MattKiazyk that's the exact command that failed to me so I guess I just need to wait...

MattKiazyk:
Thanks. I'll have to update a Mac here I guess.

RobotLeopard86:
> You should use `brew install xcodesorg/made/xcodes`
[...]

PR #816, opened 2026-06-09
Fix: macOS 27 beta launch crash

tylertms:
Issue: `AppState.init` loads cached available Xcodes, which assigns `availableXcodes`. That observer was doing installed-Xcode discovery synchronously on the main thread. On macOS 27 beta, that blocks inside `waitUntilExit()` while SwiftUI/AttributeGraph is creating `@StateObject AppState`, causing the launch crash.

#815

Fix: remove installed-Xcode discovery from the `availableXcodes` / `selectedXcodePath` observers. They now use cached installed-Xcode state only. Installed Xcodes are refreshed asynchronously after launch and before update/install resolution; install resolution refreshes without recomposing `allXcodes` so progress state is preserved.

### DIFF
--- a/Xcodes/Backend/AppState+Install.swift
+++ b/Xcodes/Backend/AppState+Install.swift
@@ -77,13 +77,15 @@ extension AppState {
     }
 
     private func getXcodeArchiveAsync(_ installationType: InstallationType, downloader: Downloader) async throws -> (AvailableXcode, URL) {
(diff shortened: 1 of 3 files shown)


tylertms:
I'm not able to test this on macOS 26, but all tests and usage pass on 27. The fix should be applicable to both versions, but it only surfaced as an issue now due to the synchronous stalling change.

PR #820, opened 2026-06-12
Fix 2FA code paste and macOS autofill in PIN code input

### LINKED ISSUE #587: Pasting 2FA SMS codes will not fill all the textfields
**Describe the bug**
Pasting 2fa codes should autofill all textfields

**To Reproduce**
1. Login using your Apple id with 2fa activated
2. Press send SMS
3. In the modal with the 6 textfields press "cmd + P" in any to paste like any other text
4. Observe how only the last digit is used only on the focused textfiled

**Expected behavior**
-

**Screenshots**
-

**Version**
 - OS: All
 - Xcodes: All

### LINKED ISSUE #788: SMS Autofill doesn't work
**Describe the bug**
When using SMS 2FA to authenticate an account, the window to enter the 6-digit code has 6 distinct text boxes, which breaks the autofill built into macOS when using 2FA codes.

**To Reproduce**
1. Authenticate with SMS text message
2. Receive a code
3. In text box to enter the code, see an auto-suggestion from macOS to fill code
4. Click the auto-suggestion
5. The first digit is entered, but not the other 5, and the autofill form dismisses, and the text message is automatically deleted.
6. Have to get a NEW code, then type it in one number at a time.

**Expected behavior**
If there were a SINGLE text box, instead of 6 individual ones, users could fill in the autofill SMS code with a single click.

**Screenshots**

**Version**
 - OS: macOS 26.2
 - Xcodes: 3.0.2 (34)

rursache:
## Problem
The 2FA/SMS code entry uses six independently focused text fields. Pasting a code only kept one digit (#587) and macOS one-time code autofill from Messages/Mail broke after the first digit because the system inserts the whole code into the focused field and the focus jumping between fields aborts the insertion (#788).

## Fix
Rewrote `PinCodeTextView` around the standard one-time code pattern: a single invisible `NSTextField` with `contentType = .oneTimeCode` overlays the row and owns all input, while the digit boxes are display-only. Pasting now works from any state, the system autofill suggestion appears (field auto-focuses when the sheet opens) and inserts the full code, and pasting a full code over partially entered digits replaces them. The public `PinCodeTextField` SwiftUI API is unchanged.

Note: the SMS being deleted after autofill is the system "Clean Up Automatically" feature and not controllable by the app.

## Testing
Verified with a headless harness driving the real field editor: paste into empty field, paste over partial input, paste with separators, digit by digit typing, backspace, overflow capping, autofill-style bulk `insertText` and external reset via the binding. App builds and runs.

Fixes #587
Fixes #788

### DIFF
--- a/Xcodes/Frontend/SignIn/PinCodeTextView.swift
+++ b/Xcodes/Frontend/SignIn/PinCodeTextView.swift
@@ -14,16 +14,9 @@ struct PinCodeTextField: NSViewRepresentable {
         view.codeDidComplete = { complete($0) }
         return view
     }
-    
+
     func updateNSView(_ nsView: NSViewType, context: Context) {
-        nsView.code = (0..<numberOfDigits).map { index in
-            if index < code.count {
-                let codeIndex = code.index(code.startIndex, offsetBy: index)
-                return code[codeIndex]
-            } else {
-                return nil
-            }
-        }
+        nsView.setCode(code)
     }
 }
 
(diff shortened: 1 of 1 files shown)


MattKiazyk:
Hi @rursache 

Was this not fixed via https://github.com/XcodesOrg/XcodesApp/pull/778 that is available in the latest 4.0.3 release?

rursache:
just had the same issue but indeed i'm on 4.0.2 not 0.3. will check again and close the PR if that the case

PR #839, opened 2026-08-05
Support password manager autofill during Apple ID sign-in

zeromhz:
Makes it possible for the password manager to autofill during Apple ID sign-ins


MattKiazyk:
Thanks for this!

PR #840, opened 2026-08-07
fix: duplicated major version in install notification title

YuriNachos:
## Problem

The macOS notification posted when an Xcode install step changes showed the major version **twice** in its title: `"\(major).\(appleDescription)"` — e.g. `27.27.0 Beta 4` for Xcode 27.0 Beta 4. `appleDescription` already starts with the major version, so prefixing `"\(major)."` duplicated it.

## Fix

Use `version.appleDescription` directly (extracted into a small testable `AppState.installNotificationTitle(for:)` helper). Xcode 27.0 Beta 4 now shows `27.0 Beta 4`; Xcode 26.5 shows `26.5`.

## Test

`test_InstallNotificationTitle_DoesNotDuplicateMajorVersion` covers a prerelease (`27.0 Beta 4`), a minor release (`26.5`), and a patch release (`10.2.1`).

PR #842, opened 2026-08-08
fix: notify on new Xcode version identity, not array-count growth

YuriNachos:
## Problem

The "New Xcode version available" notification fired whenever the `availableXcodes` array **count** grew (`newValue.count > availableXcodes.count`), not when a genuinely new version identity appeared. This caused two user-visible bugs:

- **False negative** — when a new version appears in the same refresh that an old one is removed (e.g. a data source drops an obsolete beta row the moment it adds the new release), the count is unchanged, so the user is **never told** a new Xcode landed (the core reason to run the app).
- **False positive** — a data-source switch or a duplicate row grows the count without a new version identity, firing a **spurious** banner.

## Solution

Extract a pure helper that compares **version identity**, not array length:

```swift
static func newlyAvailableXcodes(oldXcodes: [AvailableXcode], newXcodes: [AvailableXcode]) -> [AvailableXcode] {
    guard !oldXcodes.isEmpty else { return [] }
    let oldIDs = Set(oldXcodes.map(\.xcodeID))
    return newXcodes.filter { !oldIDs.contains($0.xcodeID) }
}
```

`availableXcodes.willSet` now notifies only when the helper's result is non-empty:

```swift
if !Self.newlyAvailableXcodes(oldXcodes: availableXcodes, newXcodes: newValue).isEmpty { ... }
```

- `AvailableXcode.xcodeID` (already `Hashable`) is the identity, so the same version with a different architecture correctly stays distinct and does NOT re-notify.
- The initial load (empty → populated) is suppressed: `oldXcodes.isEmpty` → `[]`, so first-run does not banner every Xcode.
- No `Localizable.xcstrings` change (title/body keys unchanged).

## Testing

New `XcodesTests/NewVersionNotificationTests.swift` (own file to avoid the `AppStateTests.swift` conflict with PR #840) asserts directly on the pure helper:

- initial-load suppression (empty → populated = `[]`);
- true positive (one genuinely new version);
- **false negative** (new added + old removed, count unchanged → the new version IS reported);
- false-negative variant (new appears while list shrinks);
- **false positive** (`[A]` → `[A, A]` → `[]`, no spurious notify);
- identical list (no notify);
- architecture-distinct same version (no re-notify).

Real red-before-green: on `main`'s count predicate, the equal-count cases (new+removed) notify nothing, so the test expecting the new version's identity fails; after the fix it passes.

`xcodebuild test -scheme Xcodes -destination 'platform=macOS' CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO -only-testing:XcodesTests/NewVersionNotificationTests` → **TEST SUCCEEDED**.

## Checklist

- [x] Diff scoped to `AppState.swift` (predicate) + new `NewVersionNotificationTests.swift` + `project.pbxproj` (test-target registration only).
- [x] No overlap with PR #840 (`AppState+Install.swift`) or #841 (`XcodeCommands.swift`).
- [x] `Localizable.xcstrings` byte-identical to `main`.
- [x] No new compiler warnings.


MattKiazyk:
Thanks for this!

